PR #1, opened 2017-10-27
Initial basic implementation

asgrim:


src/Comparator.php
```diff
@@ -0,0 +1,58 @@
...
+use Roave\BetterReflection\Reflector\ClassReflector;
+use Roave\BetterReflection\Reflector\Exception\IdentifierNotFound;
+
+final class Comparator
+{
+    public function compare(ClassReflector $oldApi, ClassReflector $newApi): array
+    {
+        $changelog = [];
+
+        foreach ($oldApi->getAllClasses() as $oldClass) {
```
Ocramius: To be isolated into a method first

src/Comparator.php
```diff
@@ -0,0 +1,58 @@
...
+            try {
+                $newClass = $newApi->reflect($oldClass->getName());
```
Ocramius: This approach only looks for BC breaks, not API additions

src/Comparator.php
```diff
@@ -0,0 +1,58 @@
...
+            } catch (IdentifierNotFound $exception) {
+                $changelog[] = sprintf('[BC] Class %s has been deleted', $oldClass->getName());
+                continue;
+            }
+
+            foreach ($oldClass->getMethods() as $oldMethod) {
```
Ocramius: To be isolated into a private method

src/Comparator.php
```diff
@@ -0,0 +1,58 @@
...
+                } catch (\OutOfBoundsException $exception) {
+                    $changelog[] = sprintf(
+                        '[BC] Method %s in class %s has been deleted',
+                        $oldMethod->getName(),
+                        $oldClass->getName()
+                    );
+                    continue;
+                }
+
+                foreach ($oldMethod->getParameters() as $oldParameter) {
```
Ocramius: And again

asgrim:
Jeez man, gimme a break, I literally just started this :P

src/Comparator.php
```diff
@@ -0,0 +1,58 @@
...
+                    $newParameter = $newMethod->getParameter($oldParameter->getName());
```
Ocramius: The parameter should not be fetched by name, but by position. A simple test of a BC compliant change is:

before:
```php
function foo(int $a, string $b) {}
```

after:
```php
function foo(int $b, string $a) {}
```

Also, a rename per-se does not cause a BC break:


before:
```php
function foo(int $a, string $b) {}
```

after:
```php
function foo(int $a, string $c) {}
```

Ocramius:
> gimme a break
rageface

src/Comparator.php
samdark: As I understand, this one is to look for public API removal, see `catch`.

asgrim:
@samdark thanks for your comments; it's still a little way off yet, hopefully I'll find time soon to work on this again :)

src/Comparator.php
```diff
@@ -0,0 +1,104 @@
...
+
+use Roave\BetterReflection\Reflection\ReflectionClass;
+use Roave\BetterReflection\Reflection\ReflectionMethod;
+use Roave\BetterReflection\Reflection\ReflectionParameter;
+use Roave\BetterReflection\Reflector\ClassReflector;
+use Roave\BetterReflection\Reflector\Exception\IdentifierNotFound;
+
+final class Comparator
+{
+    public function compare(ClassReflector $oldApi, ClassReflector $newApi): array
```
Ocramius: Do we only compare classes?

src/Comparator.php
```diff
@@ -0,0 +1,104 @@
...
+        $changelog = [];
+
+        foreach ($oldApi->getAllClasses() as $oldClass) {
+            $changelog = $this->examineClass($changelog, $oldClass, $newApi);
+        }
+
+        return $changelog;
+    }
+
+    private function examineClass(array $changelog, ReflectionClass $oldClass, ClassReflector $newApi): array
```
Ocramius: Can we avoid returning `array`?

src/Comparator.php
Ocramius: Return type `array` is too limited

src/Comparator.php
```diff
@@ -0,0 +1,104 @@
...
+
+        return $changelog;
+    }
+
+    private function examineMethod(
+        array $changelog,
+        ReflectionClass $oldClass,
+        ReflectionMethod $oldMethod,
+        ReflectionClass $newClass
+    ): array {
```
Ocramius: Same here: `array` is too limited

src/Comparator.php
asgrim: For now yes, if you want to support functions, add an issue :P

asgrim:
Created #13 to discuss the actual API to be used further :+1: for now merging so other functionality can start to be added

PR #35, opened 2018-04-15
Added a basic Markdown output formatter - and some basic docs

### LINKED ISSUE #20: Provide markdown output generation
As per #13 mwop mentioned markdown output would be useful for adding to tag messages/changlogs etc.

asgrim:
Fixes #20


README.md
```diff
@@ -1,7 +1,64 @@
...
-A tool that can be used to compare two versions of a class API.
+A tool that can be used to compare two versions of a class API in PHP
+code.
 
-## Installation & Usage 
+## Pre-requisites/assumptions
 
-TBD - still in development
+ * Your project uses `git`
+ * You use semver formatted `git` tags like `1.2.3` to mark releases
```
Ocramius: Do we support parsing things like `release-1.2.3` or `v1.2.3`?

README.md
```diff
@@ -1,7 +1,64 @@
...
+
+## Installation
+
+You must currently install the tool using:
+
+```bash
+$ composer require roave/api-compare
```
Ocramius: `composer require --dev`

README.md
```diff
@@ -1,7 +1,64 @@
...
+The typical intended usage is to just add an execution of `api-compare`
+into your CI build, by running:
+
+```bash
+$ vendor/bin/api-compare
+```
+
+This will automatically detect the last minor version tagged, and
+compare the API against the current `HEAD`. If any BC breaks are found,
+the tool returns a non-zero status, which on most CI systems will cause
```
Ocramius: Should probably document that it is `2`

README.md
```diff
@@ -1,7 +1,64 @@
...
+the build to fail.
+
+### Running manually
+
+You can also run the tool by hand, for example to generate additional
+documentation for changelogs:
+
+```bash
+$ vendor/bin/api-compare --markdown=bc-breaks.md
```
Ocramius: Let's please just use `STDOUT`. No file writing, please @_@

README.md
```diff
@@ -1,7 +1,64 @@
...
+```
+
+### CLI options
+
+Running the tool with options:
+
+```bash
+$ vendor/bin/api-compare [--from=] [--to=] [--markdown=] [<sources-path>]
```
Ocramius: Should we just call it `--format=markdown` instead?

README.md
```diff
@@ -1,7 +1,64 @@
...
+```
+
+ * `--from=<revision>` specify manually what the "old" version is (e.g.
+   `1.0.0` or a specific `git` hash). If not provided, the tool will
+   attempt to figure out the last minor version released.
+ * `--to=<revision>` specify manually what the "new" version (e.g.
+   `1.1.0` or a specific `git` hash). The default value is `HEAD`.
+ * `--markdown=<filename>` If provided, the tool will generate the list
```
Ocramius: Again: file output should probably be avoided. Better to send everything to `STDOUT`

README.md
```diff
@@ -1,7 +1,64 @@
...
+   of changes in markdown format.
+ * `<sources-path>` if given, you can specify in which directory to
+   examine for classes. This defaults to `src`.
+
+## Configuration
```
Ocramius: Drop zis

src/Command/ApiCompare.php
```diff
@@ -135,6 +137,13 @@ public function execute(InputInterface $input, OutputInterface $output) : int
             );
 
             (new SymfonyConsoleTextFormatter($output))->write($changes);
+
+            $markdownFile = trim((string)$input->getOption('markdown'));
+            if ($markdownFile !== '') {
+                $output->write(sprintf('Generating markdown in %s...', $markdownFile));
```
Ocramius: We should write to `STDOUT` and `STDERR` separately: https://gist.github.com/MattKetmo/3665488

src/Formatter/MarkdownFormatter.php
```diff
@@ -0,0 +1,60 @@
...
+     * @var string
+     */
+    private $outputFilename;
+
+    public function __construct(string $outputFilename)
+    {
+        $this->outputFilename = $outputFilename;
+    }
+
+    public function write(Changes $changes) : void
```
Ocramius: ` : string`

src/Formatter/MarkdownFormatter.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $arrayOfChanges = $changes->getIterator()->getArrayCopy();
+
+        file_put_contents(
```
Ocramius: As of above: to be avoided. Let's rely on standard piping and output redirection techniques instead.

README.md
asgrim: For now at least, no. Issue maybe?

README.md
asgrim: Why?

Ocramius: Bcoz empty

Ocramius: > There are currently no configuration options available.

PR #38, opened 2018-04-15
Verify method, function, property, constant and interface changes and report BC breaks

### LINKED ISSUE #7: Identify added/changed parameter type declarations for public/protected methods
`class A { public function a($a) {} }`
`class A { public function a(int $a) {} }`

`class A { public function a(int $a) {} }`
`class A { public function a(string $a) {} }`

`class A { protected function a($a) {} }`
`class A { protected function a(int $a) {} }`

`class A { protected function a(int $a) {} }`
`class A { protected function a(string $a) {} }`

### LINKED ISSUE #8: Identify added return type to a protected/public method
`class A { public function a() {} }`
`class A { public function a(): int {} }`

`class A { protected function a() {} }`
`class A { protected function a(): int {} }`

but *NOT* (possibly?)

`final class A { function a() {} }`
`final class A { function a(): int {} }`

### LINKED ISSUE #9: Identify changing static flag on a method
`class A { public function a() {} }`
`class A { public static function a() {} }`

`class A { public static function a() {} }`
`class A { public function a() {} }`

### LINKED ISSUE #10: Identify removed public/protected class constants
PHP 7.2 only:
`class A { public constant A = ''; }`
`class A { }`

PHP 7.2 only:
`class A { protected constant A = ''; }`
`class A { }`

PHP 7.2 only, and *NOT*:
`class A { private constant A = ''; }`
`class A { }`

Any PHP version:
`class A { constant A = ''; }`
`class A { }`

### LINKED ISSUE #11: Identify changed visibility of constants
PHP 7.2 only:
`class A { public constant A = ''; }`
`class A { protected constant A = ''; }`

PHP 7.2 only:
`class A { protected constant A = ''; }`
`class A { private constant A = ''; }`

PHP 7.2 only, and *NOT*:
`class A { public constant A = ''; }`
`class A { private constant A = ''; }`

### LINKED ISSUE #12: Identify added parameters to methods
`class A { public function a() {} }`
`class A { public function a(int $a) {} }`

`class A { protected function a() {} }`
`class A { protected function a(int $a) {} }`

`class A { public function a() {} }`
`class A { public function a(int $a = 1) {} }`

`class A { protected function a() {} }`
`class A { protected function a(int $a = 1) {} }`

`final class A { public function a() {} }`
`final class A { public function a(int $a) {} }`

`final class A { protected function a() {} }`
`final class A { protected function a(int $a) {} }`

but *NOT*:
`final class A { public function a() {} }`
`final class A { public function a(int $a = 1) {} }`

and *NOT*:
`final class A { protected function a() {} }`
`final class A { protected function a(int $a = 1) {} }`

### LINKED ISSUE #41: identify changes to references on methods/parameters
_yes, I really said that ◉_◉_

### LINKED ISSUE #43: Design trait change BC detectors
As per https://github.com/Roave/ApiCompare/pull/38#issuecomment-381790429, we need to also work on `trait`. This is low priority for now, but it will eventually come up in the next major (after `1.0.0`)

### LINKED ISSUE #46: Detect changes in parameter and return type declarations of open classes and interfaces as BC breaks
https://github.com/Roave/ApiCompare/pull/38#discussion_r182247543

Ocramius:
Fixes #7
Fixes #8
Fixes #9
Fixes #10
Fixes #11
Fixes #12
Fixes #41 
Fixes #43
Fixes #46

- [x] requires https://github.com/Roave/BetterReflection/pull/415
- [x] class became interface
- [x] class became abstract
- [x] interface became final/regular class
- [x] interface became abstract class
- [x] method became final
- [x] method became abstract


composer.json
```diff
@@ -4,7 +4,7 @@
     "require": {
         "php": "^7.2",
         "roave/security-advisories": "dev-master",
-        "roave/better-reflection": "^2.0",
+        "roave/better-reflection": "dev-master#583ce1321dc7d5987fcd535230edcb0960c2d6c6",
```
Ocramius: - [ ] To be changed before merging

Ocramius:
Overall awesome list, but this one shouldn't be touched at all:

 >  how to make traits fit whole concept? they are definíitely types, but there are weird use cases where use trait X for Y is suggested

Traits are not types: they integrate with concrete types once used. The BC breaks on the public API of a trait are most likely the same as the ones in a concrete class, except that the breakages also apply to `private` members.

 > Supporting phpDoc for specializing types, like arrays, iterables etc. Even though signature itself stays unchanged, array contents may change in an incompatible way.

This one I skipped mostly due to complexity, but already implemented something simplistic for the properties, so the implementation path is quite clear unless we need to import a library just to couple `@param` with a parameter :-P

Majkl578:
> Traits are not types

Obviously, **_not_** got lost somehow. :sob: My point is that traits are sometimes misused as types (i.e. declare behavior + abstract methods).

Ocramius:
I'm going to open a separate issue for traits, because:

 1. it is unclear how they should be treated
 2. they are an abomination overall
 3. I CBA to fix it now, and would rather see the important bits running first :+1:

Ocramius:
https://github.com/Roave/ApiCompare/issues/43

Ocramius:
https://github.com/Roave/ApiCompare/issues/44 for the `@param` docblocks

Ocramius:
https://github.com/Roave/ApiCompare/issues/45 is for the `@return` docblocks

Ocramius:
>  extended constant value change detection (i.e. is expanding an array with new value a BC break?)

No, not a BC break. This is more like a coding principles violation, but here we are talking about changing a constant value (already covered by `ConstantBased` BC checks)

src/Comparator/BackwardsCompatibility/FunctionBased/ParameterTypeContravarianceChanged.php
```diff
@@ -0,0 +1,88 @@
...
+        $toType   = $toParameter->getType();
+
+        if ($this->typeIsContravariant->__invoke($fromType, $toType)) {
+            return Changes::new();
+        }
+
+        return Changes::fromArray([
+            Change::changed(
+                sprintf(
+                    'The parameter $%s of %s() changed from %s to a non-contravariant %s',
```
asgrim: Wonder if we can word better for people who don't understand covariance/contravariance?

src/Comparator/BackwardsCompatibility/FunctionBased/ReturnTypeCovarianceChanged.php
```diff
@@ -0,0 +1,71 @@
...
+        $toReturnType   = $toFunction->getReturnType();
+
+        if ($this->typeIsCovariant->__invoke($fromReturnType, $toReturnType)) {
+            return Changes::new();
+        }
+
+        return Changes::fromArray([
+            Change::changed(
+                sprintf(
+                    'The return type of %s changed from %s to the non-covariant %s',
```
asgrim: Again wording for people who don't understand covariance/contravariance

src/Comparator/BackwardsCompatibility/MethodBased/MethodScopeChanged.php
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * A method that changes from instance to static or the opposite has to be called differently,
+ * so any of such changes are to be considered BC breaks
+ */
+final class MethodScopeChanged implements MethodBased
+{
+    public function compare(ReflectionMethod $fromMethod, ReflectionMethod $toMethod) : Changes
+    {
+        if ($fromMethod->isPrivate()) {
```
asgrim: Didn't this already get extracted in `AccessibleMethodFunctionBasedChange`?

src/Comparator/BackwardsCompatibility/PropertyBased/PropertyDefaultValueChanged.php
```diff
@@ -0,0 +1,39 @@
...
+
+use Roave\ApiCompare\Change;
+use Roave\ApiCompare\Changes;
+use Roave\BetterReflection\Reflection\ReflectionProperty;
+
+final class PropertyDefaultValueChanged implements PropertyBased
+{
+    public function compare(ReflectionProperty $fromProperty, ReflectionProperty $toProperty) : Changes
+    {
+        if ($fromProperty->isPrivate()) {
```
asgrim: Can this be extracted like `AccessibleMethodFunctionBasedChange` ?

src/Comparator/BackwardsCompatibility/PropertyBased/PropertyDocumentedTypeChanged.php
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Type declarations for properties are invariant: you can't restrict the type because the consumer may
+ * write invalid values to it, and you cannot widen the type because the consumer may expect a specific
+ * type when reading.
+ */
+final class PropertyDocumentedTypeChanged implements PropertyBased
+{
+    public function compare(ReflectionProperty $fromProperty, ReflectionProperty $toProperty) : Changes
+    {
+        if ($fromProperty->isPrivate()) {
```
asgrim: Can this be extracted with `AccessibleMethodFunctionBasedChange`?

src/Comparator/BackwardsCompatibility/PropertyBased/PropertyDocumentedTypeChanged.php
```diff
@@ -0,0 +1,51 @@
...
+use Roave\ApiCompare\Change;
+use Roave\ApiCompare\Changes;
+use Roave\BetterReflection\Reflection\ReflectionProperty;
+
+/**
+ * Type declarations for properties are invariant: you can't restrict the type because the consumer may
+ * write invalid values to it, and you cannot widen the type because the consumer may expect a specific
+ * type when reading.
+ */
+final class PropertyDocumentedTypeChanged implements PropertyBased
```
asgrim: Hmm, not sure about using docblocks to trigger BC breaks tbh...

guidelines

src/Comparator/BackwardsCompatibility/PropertyBased/PropertyVisibilityReduced.php
```diff
@@ -0,0 +1,53 @@
...
+    private const VISIBILITY_PROTECTED = 'protected';
+
+    private const VISIBILITY_PUBLIC = 'public';
+
+    public function compare(ReflectionProperty $fromProperty, ReflectionProperty $toProperty) : Changes
+    {
+        $visibilityFrom = $this->propertyVisibility($fromProperty);
+        $visibilityTo   = $this->propertyVisibility($toProperty);
+
+        if ($visibilityFrom <= $visibilityTo) {
```
asgrim: Something something sorting comment

src/Comparator/BackwardsCompatibility/PropertyBased/PropertyVisibilityReduced.php
```diff
@@ -0,0 +1,53 @@
...
+            return Changes::new();
+        }
+
+        return Changes::fromArray([Change::changed(
+            sprintf(
+                'Property %s#$%s visibility reduced from %s to %s',
```
asgrim: Just a minor thing, but do we have consistency with using `#` vs `::`? What are the rules here for formatting?

src/Comparator/Variance/TypeIsContravariant.php
```diff
@@ -0,0 +1,74 @@
...
+
+use Roave\BetterReflection\Reflection\ReflectionType;
+use function in_array;
+use function strtolower;
+
+/**
+ * This is a simplistic contravariant type check. A more appropriate approach would be to
+ * have a `$type->includes($otherType)` check with actual types represented as value objects,
+ * but that is a massive piece of work that should be done by importing an external library
+ * instead, if this class no longer suffices.
```
asgrim: Issue for this?

bin/api-compare.php
```diff
@@ -31,7 +39,44 @@
             new GetVersionCollectionFromGitRepository(),
             new PickLastMinorVersionFromCollection(),
             new Comparator(
-                new Comparator\BackwardsCompatibility\ClassBased\PropertyRemoved()
+                new ClassBased\MultiClassBased(
```
asgrim: Could probably need documentation (in `docs/` folder) explaining each of this; it's fine to put in a new issue for now

.travis.yml
```diff
@@ -3,6 +3,10 @@ language: php
 before_script:
   - composer install
 
+# see https://github.com/travis-ci/travis-ci/issues/4942#issuecomment-159132444
+git:
+  depth: 99999
```
asgrim: Why do we need unshallow clones? I don't see anything requiring it?

src/Comparator/BackwardsCompatibility/FunctionBased/ParameterTypeContravarianceChanged.php
```diff
@@ -0,0 +1,88 @@
...
+use Roave\ApiCompare\Change;
+use Roave\ApiCompare\Changes;
+use Roave\ApiCompare\Comparator\Variance\TypeIsContravariant;
+use Roave\BetterReflection\Reflection\ReflectionFunctionAbstract;
+use Roave\BetterReflection\Reflection\ReflectionMethod;
+use Roave\BetterReflection\Reflection\ReflectionParameter;
+use Roave\BetterReflection\Reflection\ReflectionType;
+
+/**
+ * When a parameter type changes, the new type should be wider than the previous type, or else
```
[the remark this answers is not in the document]
Ocramius: These are `FunctionBased` checks: the check only verifies BC from a consumer perspective, not from an inheritance perspective.I think I didn't cover the signature changes (for subclassing) at all, but I'll open a separate issue (https://github.com/Roave/ApiCompare/issues/46)

src/Comparator/BackwardsCompatibility/MethodBased/MethodBecameFinal.php
```diff
@@ -0,0 +1,33 @@
...
+namespace Roave\ApiCompare\Comparator\BackwardsCompatibility\MethodBased;
+
+use Roave\ApiCompare\Change;
+use Roave\ApiCompare\Changes;
+use Roave\BetterReflection\Reflection\ReflectionMethod;
+use function sprintf;
+
+/**
+ * A method that changes from non-final to final breaks all child classes that
+ * override it.
```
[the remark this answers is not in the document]
Ocramius: Indeed, I didn't touch the `final` stuff yet. Will likely make a filter for it, much like I did for applying `FunctionBased` checks to `MethodBased` checks. https://github.com/Roave/ApiCompare/issues/31

src/Comparator/BackwardsCompatibility/FunctionBased/ParameterTypeContravarianceChanged.php
Ocramius: covariance/contravariance are the correct terms here - I wouldn't invent my own or pick something more vague. Covariance/contravariance are a "must know" for library developers.

src/Comparator/BackwardsCompatibility/MethodBased/MethodScopeChanged.php
Ocramius: Same as https://github.com/Roave/ApiCompare/pull/38#discussion_r182383370 - covariance/contravariance are concepts that need to be known/learned: no escape from that.

src/Comparator/BackwardsCompatibility/PropertyBased/PropertyDefaultValueChanged.php
Ocramius: Most likely, yes. And good call too :+1:

src/Comparator/BackwardsCompatibility/PropertyBased/PropertyDocumentedTypeChanged.php
Ocramius: Property docblocks are the only guarantee (although indeed really weak) that we have for now.

As I documented in #44 and #45, we still need to add them to methods too.

PHP is a shitty language, so the documented types are still the best level of "trust" that we can put in an API.

src/Comparator/BackwardsCompatibility/PropertyBased/PropertyVisibilityReduced.php
Ocramius: I'm trying to use `#` for instance, `::` for statics. I'll extract this to a helper, much like for the function naming.

src/Comparator/Variance/TypeIsContravariant.php
Ocramius: Not really: I think we're fine with this for now, but we will hit edge cases when we start supporting union, intersection and nested types (`(Foo[]|Bar)[][]` for example) in #44 and #45. Luckily, property types are invariant, so any type change is a BC break :D

bin/api-compare.php
Ocramius: Would probably just document each class, no?

If they are unclear, then I suggest correcting the naming instead

.travis.yml
Ocramius: Somewhere in the CI build you are running these checks against the library itself, and this patch already went further than a shallow clone allows.

src/Comparator/BackwardsCompatibility/MethodBased/MethodScopeChanged.php
Ocramius: Interestingly, this is extremely useful when I'm going to introduce a `TraitBased` comparator, because the same checks can now be used both for checking `private` and `protected`/`public` changes.

I made an `AccessiblePropertyChanged` for that for now :+1:

src/Comparator/BackwardsCompatibility/FunctionBased/ReturnTypeCovarianceChanged.php
Ocramius: Same as https://github.com/Roave/ApiCompare/pull/38#discussion_r182383370 - covariance/contravariance are concepts that need to be known/learned: no escape from that.

Ocramius:
Just tried this against this library directly:

```
./bin/api-compare --from=dd5cabd1408bab2b3db2b6d6e9cb608db5661540 --to=HEAD
Comparing from dd5cabd1408bab2b3db2b6d6e9cb608db5661540 to ce49948483e2b2612a8a21a6cf22c554476e270b...
[BC] CHANGED: The parameter $functionBasedComparisons of Roave\ApiCompare\Comparator#__construct() changed from Roave\ApiCompare\Comparator\BackwardsCompatibility\FunctionBased\FunctionBased to a non-contravariant Roave\ApiCompare\Comparator\BackwardsCompatibility\InterfaceBased\InterfaceBased
[BC] CHANGED: The parameter $functionBasedComparisons of Roave\ApiCompare\Comparator#__construct() changed from Roave\ApiCompare\Comparator\BackwardsCompatibility\FunctionBased\FunctionBased to Roave\ApiCompare\Comparator\BackwardsCompatibility\InterfaceBased\InterfaceBased
[BC] REMOVED: Class Roave\ApiCompare\Comparator\BackwardsCompatibility\ClassBased\PropertyVisibilityReduced has been deleted
[BC] REMOVED: Class Roave\ApiCompare\Comparator\BackwardsCompatibility\ClassBased\MethodVisibilityReduced has been deleted
[BC] REMOVED: Class Roave\ApiCompare\Comparator\BackwardsCompatibility\ClassBased\ConstantValueChanged has been deleted
[BC] REMOVED: Class Roave\ApiCompare\Comparator\BackwardsCompatibility\ClassBased\ConstantVisibilityReduced has been deleted
[BC] REMOVED: Class Roave\ApiCompare\Comparator\BackwardsCompatibility\ClassBased\MethodScopeChanged has been deleted
```

Ocramius:
@asgrim this is ready for another round of reviews

asgrim:
@Ocramius nice - sorry I didn't get to this today, will try and look soon! :+1:

phpcs.xml.dist
```diff
@@ -12,6 +16,19 @@
     <file>./src</file>
     <file>./test/unit</file>
 
+    <rule ref="Doctrine">
```
asgrim: Why are we excluding all these rules, mainly out of interest...?

src/Comparator/BackwardsCompatibility/ClassBased/MethodChanged.php
```diff
@@ -0,0 +1,58 @@
...
+use function array_map;
+use function array_reduce;
+use function strtolower;
+
+final class MethodChanged implements ClassBased
+{
+    /** @var MethodBased */
+    private $checkMethod;
+
+    public function __construct(MethodBased $checkMethod)
```
asgrim: I'm liking this new nested structure! A little more in depth to start with, but it makes a lot of sense :+1:

asgrim: Incidentally; could `ClassBased` be a "child" of `InterfaceBased`, considering a class is a specialisation of an interface, perhaps?

src/Comparator/BackwardsCompatibility/ClassBased/MethodRemoved.php
```diff
@@ -20,6 +21,14 @@
 
 final class MethodRemoved implements ClassBased
 {
+    /** @var ReflectionFunctionAbstractName */
+    private $formatFunction;
+
+    public function __construct()
+    {
+        $this->formatFunction = new ReflectionFunctionAbstractName();
```
asgrim: :+1:

src/Comparator/BackwardsCompatibility/ClassConstantBased/ConstantVisibilityReduced.php
```diff
@@ -0,0 +1,52 @@
...
+    private const VISIBILITY_PROTECTED = 'protected';
+
+    private const VISIBILITY_PUBLIC = 'public';
+
+    public function compare(ReflectionClassConstant $fromConstant, ReflectionClassConstant $toConstant) : Changes
+    {
+        $visibilityFrom = $this->propertyVisibility($fromConstant);
+        $visibilityTo   = $this->propertyVisibility($toConstant);
+
+        if ($visibilityFrom <= $visibilityTo) {
```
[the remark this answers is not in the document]
asgrim: Fixed :+1:

src/Comparator/BackwardsCompatibility/FunctionBased/ParameterTypeContravarianceChanged.php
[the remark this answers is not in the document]
asgrim: ok :+1:

src/Comparator/BackwardsCompatibility/InterfaceBased/InterfaceBecameClass.php
```diff
@@ -27,6 +28,7 @@ public function compare(ReflectionClass $fromClass, ReflectionClass $toClass) :
         return Changes::fromArray([Change::changed(
             sprintf('Interface %s became a class', $fromClass->getName()),
             true
-        )]);
+        ),
```
asgrim: not sure what's going on here

src/Comparator/BackwardsCompatibility/MethodBased/MethodBecameFinal.php
```diff
@@ -28,6 +28,7 @@ public function compare(ReflectionMethod $fromMethod, ReflectionMethod $toMethod
                 $fromMethod->getDeclaringClass()->getName()
             ),
             true
-        )]);
+        ),
```
asgrim: This seems like such a pointless CS change....................?

src/Comparator/BackwardsCompatibility/MethodBased/MethodVisibilityReduced.php
```diff
@@ -0,0 +1,53 @@
...
+    private const VISIBILITY_PROTECTED = 'protected';
+
+    private const VISIBILITY_PUBLIC = 'public';
+
+    public function compare(ReflectionMethod $fromMethod, ReflectionMethod $toMethod) : Changes
+    {
+        $visibilityFrom = $this->methodVisibility($fromMethod);
+        $visibilityTo   = $this->methodVisibility($toMethod);
+
+        if ($visibilityFrom <= $visibilityTo) {
```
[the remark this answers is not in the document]
asgrim: Dealt with

src/Comparator/BackwardsCompatibility/PropertyBased/PropertyDocumentedTypeChanged.php
asgrim: Not convinced BC breaks based only on phpdocs (considering how incorrect they can be.........) is a good idea. If anything at the moment, I'd make these opt-in tbh, with maybe some configuration / param option.

(cue some stupid YAML-based joke that serves no purpose but to undermine my genuine concern for some kind of configurability)

src/Formatter/ReflectionFunctionAbstractName.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\ApiCompare\Formatter;
+
+use Roave\BetterReflection\Reflection\ReflectionFunctionAbstract;
+use Roave\BetterReflection\Reflection\ReflectionMethod;
+
+final class ReflectionFunctionAbstractName
```
asgrim: I approve :+1:

bin/api-compare.php
```diff
@@ -31,7 +40,210 @@
             new GetVersionCollectionFromGitRepository(),
             new PickLastMinorVersionFromCollection(),
             new Comparator(
-                new Comparator\BackwardsCompatibility\ClassBased\PropertyRemoved()
```
Ocramius: Note: this tree of checks is quite huge, but is pretty much describing what kinds of BC breaks we want to rely upon.

If we want to make it more "friendly", I can move the instantiation to `Roave\ApiCompare\ApiCompare#defaultClassChecks()`, `Roave\ApiCompare\ApiCompare#defaultInterfaceChecks()`, `Roave\ApiCompare\ApiCompare#defaultTraitChecks()`,

Ocramius:
@asgrim this is ready for another round. Renamed few things, introduced traits checks.

src/Comparator/BackwardsCompatibility/ClassBased/ClassBecameTrait.php
```diff
@@ -0,0 +1,34 @@
...
+        Assert::that($fromClass->getName())->same($toClass->getName());
+
+        if ($fromClass->isTrait() || ! $toClass->isTrait()) {
+            return Changes::new();
+        }
+
+        return Changes::fromArray([Change::changed(
+            sprintf('Class %s became a trait', $fromClass->getName()),
+            true
+        ),
```
asgrim: Alignment looks wonky here?

src/Comparator/BackwardsCompatibility/InterfaceBased/InterfaceBecameClass.php
```diff
@@ -31,4 +31,12 @@ public function compare(ReflectionClass $fromClass, ReflectionClass $toClass) :
         ),
         ]);
     }
+
+    /**
+     * According to the current state of the PHP ecosystem, we only have traits, interfaces and classes
+     */
+    private function isClass(ReflectionClass $class) : bool
```
asgrim: Does `ReflectionClass` really not have an `isClass` method... seems like an oversight >.<

src/Comparator/BackwardsCompatibility/InterfaceBased/InterfaceBecameTrait.php
```diff
@@ -0,0 +1,34 @@
...
+
+        if (! $toClass->isTrait() || ! $fromClass->isInterface()) {
+            // checking whether an interface became an class is done in `InterfaceBecameClass`
+            return Changes::new();
+        }
+
+        return Changes::fromArray([Change::changed(
+            sprintf('Interface %s became a trait', $fromClass->getName()),
+            true
+        ),
```
asgrim: Something weird causing these alignment issues in phpcs I expect...

src/Comparator/BackwardsCompatibility/TraitBased/TraitBecameClass.php
```diff
@@ -0,0 +1,41 @@
...
+    public function compare(ReflectionClass $fromTrait, ReflectionClass $toTrait) : Changes
+    {
+        Assert::that($fromTrait->getName())->same($toTrait->getName());
+
+        if ($this->isClass($fromTrait) || ! $this->isClass($toTrait)) {
+            return Changes::new();
+        }
+
+        return Changes::fromArray([Change::changed(
+            sprintf('Interface %s became a class', $fromTrait->getName()),
```
asgrim: `Trait %s became a class`

src/Comparator/BackwardsCompatibility/TraitBased/TraitBecameInterface.php
```diff
@@ -0,0 +1,33 @@
...
+    public function compare(ReflectionClass $fromTrait, ReflectionClass $toTrait) : Changes
+    {
+        Assert::that($fromTrait->getName())->same($toTrait->getName());
+
+        if ($toTrait->isTrait() || ! $toTrait->isInterface() || ! $fromTrait->isTrait()) {
+            return Changes::new();
+        }
+
+        return Changes::fromArray([Change::changed(
+            sprintf('Interface %s became an interface', $fromTrait->getName()),
```
asgrim: `Trait %s became an interface`

src/Support/ArrayHelpers.php
```diff
@@ -0,0 +1,29 @@
...
+     * Yes, this is just a very pedantic version of `in_array()`, written to avoid mutations and
+     * designed to throw an exception if `$arrayOfStrings` is not a `string[]` as requested.
+     *
+     * @param string[] $arrayOfStrings
+     *
+     * @throws \InvalidArgumentException
+     */
+    public static function stringArrayContainsString(string $value, array $arrayOfStrings) : bool
+    {
+        Assert::that($arrayOfStrings)->all()->string();
```
asgrim: Really not sure we need this... seems to just be avoiding false positives; don't know specifically what mutations phpstan performs, but adding an integer onto this array shouldn't be a problem if the expected value is still in there (because we (were) using "strict" mode)... Think this conversation should be moved to a discussion upstream in phpstan, not solved with a wrapper IMO

src/Command/ApiCompare.php
```diff
@@ -132,7 +129,7 @@ public function execute(InputInterface $input, OutputInterface $output) : int
             $outputFormats = $input->getOption('format') ?: [];
             Assert::that($outputFormats)->isArray();
 
-            if (in_array('markdown', $outputFormats, true)) {
+            if (ArrayHelpers::stringArrayContainsString('markdown', $outputFormats)) {
```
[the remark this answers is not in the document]
Ocramius: I want to avoid colliding with various tools that say different things here:

 * infection
 * ea inspections
 * phpstan strict checks

Segregating it out was easy peasy, so I just did it, and we have a single location where `in_array()` still exists.

Yes, this is indeed overblown, but I'd be OK with this particular addition, considering that it really is the only thing needed to make mutation tests work without problems (could've been much worse).

src/Comparator.php
```diff
@@ -19,12 +20,17 @@ class Comparator
     /** @var InterfaceBased */
     private $interfaceBasedComparisons;
 
+    /** @var TraitBased */
+    private $traitBasedComparisons;
```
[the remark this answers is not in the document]
Ocramius: Agreed - did it mostly due to time constraints on my end, as I was in a hurry/constantly on the road yesterday.

composer.json
```diff
@@ -21,7 +21,8 @@
         "phpunit/phpunit": "^7.0",
         "doctrine/coding-standard": "^4.0",
         "squizlabs/php_codesniffer": "^3.2",
-        "phpstan/phpstan": "^0.9.2"
+        "phpstan/phpstan": "^0.9.2",
+        "infection/infection": "^0.8.1"
```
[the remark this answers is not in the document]
Ocramius: Same as https://github.com/Roave/ApiCompare/pull/38#discussion_r183358196 - indeed creeped in too much

src/Comparator/BackwardsCompatibility/ClassBased/ClassBecameTrait.php
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * A class cannot become a trait without introducing an explicit BC break, since
+ * all child classes or implementors need to be changed from `extends` to `use`,
+ * and all instantiations start failing
+ */
+final class ClassBecameTrait implements ClassBased
+{
+    public function compare(ReflectionClass $fromClass, ReflectionClass $toClass) : Changes
+    {
+        Assert::that($fromClass->getName())->same($toClass->getName());
```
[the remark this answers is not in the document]
Ocramius: Aye, should probably drop these assertions from all `ClassBased` checks

src/Command/ApiCompare.php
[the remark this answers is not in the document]
asgrim: Another potential issue: contributors (me included) will likely default to just using `in_array` in future contributions, that may slip through the net :/ it's a non-standard approach, which is "unexpected" and inconsistent with current ... "ways" (for want of a better word)

src/Comparator/BackwardsCompatibility/InterfaceBased/InterfaceBecameClass.php
Ocramius: Doesn't have one, no. That's because we mimicked the existing `ext-reflection` API, without designing our own API from scratch upfront.

src/Command/ApiCompare.php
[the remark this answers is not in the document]
Ocramius: No, it won't fail with a matching error message, obviously.

 > the solution is not what people will expect. 

Agreed, but I don't expect that to be a blocker, as a quick review can deal with most of these errors.

 > the solution is unexpected and unconventional.

Agreed again, as I said, the mutation testing outcome was much better than expected, so I'm quite happy with having this deviation from the standard compared to what it could have been if we had much more mutations. 

 > apart from this shouldn't be a problem and that the finger points back to the tools (phpstan, in this specific case)...

Possible, but I don't think that should become a blocker for just a workaround in here. Assuming that we fix phpstan there, we can really just revert this addition (API is marked `@internal` anyway) and that's the end of the story :-P

src/Comparator/BackwardsCompatibility/InterfaceBased/InterfaceBecameTrait.php
Ocramius: Verified this: it's an issue in one of the sniffs.

Basically, single-value array should be on a single line, but then the value turns out to be multi-line here. Trying to fix it leads to a phpcs violation being reported, so I'll keep these wonky ones as they are here, and talk to the doctrine folks about it.

Ocramius:
@asgrim feedback incorporated into the patch, but I'll keep the `in_array()` replacement in it for now

PR #50, opened 2018-04-27
#24 rename package to `roave/backward-compatibility-check`

### LINKED ISSUE #24: Re-naming the package...?
I've noticed when explaining what the tool does, folks jump to conclusion that this is used for comparing HTTP/REST API versions. I wonder if renaming the package could alleviate this problem. Would be better to do it before we tag anything as naturally it's (ironically) a big BC break ;)

Not sure WHAT to suggest though for naming, feedback welcome on this.

Ocramius:
Fixes #24


.travis.yml
```diff
@@ -34,6 +34,10 @@ jobs:
       php: 7.2
       script: vendor/bin/infection
 
+    - stage: Verify BC Breaks
+      php: 7.2
+      script: bin/roave-backward-compatibility-check --from=master --to=HEAD
```
Ocramius: Note: this will fail for now, as this patch does indeed lead to BC breaks (namespace changes)

README.md
```diff
@@ -1,66 +1,54 @@
-# API Compare Tool
+# Roave Backward Compatibility Check
```
Ocramius: Overall, I aimed at a few things:

 1. cut to the point in the docs
 2. moved docs for the CLI command into the CLI command itself
 3. made the `roave` bit obnoxiously repetitive: I like throwing free time at OSS, but people can indeed type `roave` a few times.

src/Changes.php
```diff
@@ -2,55 +2,53 @@
...
-     * @param Change[] $changes
-     * @return Changes
-     */
-    public static function fromArray(array $changes) : self
-    {
-        Assert::that($changes)->all()->isInstanceOf(Change::class);
-        $instance          = self::new();
-        $instance->changes = $changes;
-        return $instance;
+        if ($empty) {
```
Ocramius: I found (via `Kcachegrind`) that this was really trashing memory and CPU on this side (together with `fromArray()` and `mergeWith()`)

src/Changes.php
```diff
@@ -2,55 +2,53 @@
...
         return $instance;
     }
 
-    public function withAddedChange(Change $change) : self
+    public function mergeWith(self $other) : self
     {
-        $new            = clone $this;
-        $new->changes[] = $change;
-        return $new;
+        if (! $other->changes) {
```
Ocramius: Looked useless at first, but saves us from loads of instantiations

src/DetectChanges/BCBreak/MethodBased/MethodBased.php
```diff
@@ -0,0 +1,13 @@
...
+declare(strict_types=1);
+
+namespace Roave\BackwardCompatibility\DetectChanges\BCBreak\MethodBased;
+
+use Roave\BackwardCompatibility\Changes;
+use Roave\BetterReflection\Reflection\ReflectionMethod;
+
+interface MethodBased
+{
+    public function __invoke(ReflectionMethod $fromMethod, ReflectionMethod $toMethod) : Changes;
```
Ocramius: All checkers are now functional interfaces - didn't like the repeated name, nor the `compare()` wording, and settled with using just the wording from the class name

asgrim:
Build failures were due to BC breaks indeed, so merging.

PR #59, opened 2018-05-03
Use native invocations instead of explicit __invoke()

Majkl578:
All these `->__invoke()` things are not cool when language provides native syntax...

### DIFF
--- a/src/Command/AssertBackwardsCompatible.php
+++ b/src/Command/AssertBackwardsCompatible.php
@@ -167,18 +167,18 @@ USAGE
             Assert::that($fromSources)->directory();
             Assert::that($toSources)->directory();
 
-            $changes = $this->compareApi->__invoke(
-                $this->reflectorFactory->__invoke(
+            $changes = ($this->compareApi)(
+                ($this->reflectorFactory)(
                     $fromPath . '/' . $sourcesPath,
                     new AggregateSourceLocator() // no dependencies
                 ),
-                $this->reflectorFactory->__invoke(
+                ($this->reflectorFactory)(
                     $fromPath . '/' . $sourcesPath,
-                    $this->locateDependencies->__invoke((string) $fromPath)
+                    ($this->locateDependencies)((string) $fromPath)
                 ),
-                $this->reflectorFactory->__invoke(
+                ($this->reflectorFactory)(
                     $toPath . '/' . $sourcesPath,
-                    $this->locateDependencies->__invoke((string) $toPath)
+                    ($this->locateDependencies)((string) $toPath)
                 )
             );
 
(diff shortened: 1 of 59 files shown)


Ocramius (changes requested):
The reason for the explicit invocations is that the tooling (PHPStorm, specifically) doesn't yet detect them via "find usages", so I'd rather leave them in as we heavily rely on tge IDE for refactoring purposes.

greg0ire:
IMHO the IDE should adapt to the code, not the other way around.

Majkl578:
Well, if you are a hostage of your own IDE, then something is a bit wrong...

Ocramius:
Yes, but since I heavily rely on it, I'm making a trade-off and keeping it until the issue is actually solved.

Ocramius:
Yes, I'm hostage of it: wouldn't go with VIM unless paid at my usual hourly rate to do so 😆

greg0ire:
We don't want more people, all 4 of us vimmers are just fine :P

ostrolucky:
"things are not cool" is justification of this change? really? This change has multiple drawbacks, you better provide some practical reasons for this.

Majkl578:
@Ocramius When I first saw your WIP PR, I thought _"Wow, callables, he's taking it to whole next level, chaining callables etc!"_.
I suppose I was wrong. As it is now, it's just a set of ugly underscored method calls.
Sorry for interrupting.

Ocramius:
No, it also removes the need for a method name overall, moving the problem to just to the class name.

Another thing that is also immediately clear to consumers is that this is the only public API for each class, and it prevents creeping in multiple methods per class.

Majkl578:
> it also removes the need for a method name overall

You still need to name it, be it ugly `__invoke()` or nice `invoke()`.

>  immediately clear to consumers is that this is the only public API

Not really sice you can have `__invoke()` and normal methods side by side.

> it prevents creeping in multiple methods per class

Just like any other class with any single public method.

Majkl578:
@asgrim Just out of curiosity, can I have your opinion?

Majkl578:
Reopening this as the PhpStorm bug has been fixed and it should land in next EAP. :tada:

ostrolucky:
There are deeper issues with dynamic invocation, not a minor braces bug, like https://youtrack.jetbrains.com/issue/WI-34223

You also can't ctrl+click to go to method definition, not sure how do you imagine IDE to solve that.

Majkl578:
Sounds like a good idea! 
/cc @carusogabriel @kukulich?

kukulich:
I like the idea :)

Ocramius:
BTW, PHPStorm now seems to correctly detect `__invoke` usages: issue can be re-evaluated

kukulich:
https://github.com/slevomat/coding-standard/issues/899

Ocramius:
Can be changed by adjusting `SlevomatCodingStandard.PHP.DisallowDirectMagicInvokeCall`

Ocramius:
I've rebased your commits and merged it all into #323, thanks @Majkl578!

PR #61, opened 2018-05-04
Bump nikolaposa/version version

### LINKED ISSUE #37: Allow release versions such as `v1.2.3` and `release-1.2.3` (as well as other patterns recognized by composer)
Ref: https://github.com/Roave/ApiCompare/pull/35#discussion_r181588570

ntzm:
Partially fixes #37, allows for `v` prefix but not `release-`

PHPStan is failing due to the return type for `VersionsCollection::getIterator` is `Traversable`, which doesn't have the `current` method. Not sure what the best fix is here?

Tests and code for checking if a `VersionsCollection` is empty have been removed, due to `VersionsCollection` requiring at least one `Version` in the constructor.


src/Git/GetVersionCollectionFromGitRepository.php
```diff
@@ -30,7 +30,7 @@ public function fromRepository(CheckedOutRepository $checkedOutRepository) : Ver
             ->mustRun()
             ->getOutput();
 
-        return VersionsCollection::fromArray(array_filter(
+        return new VersionsCollection(...array_filter(
```
Ocramius: An `array_values()` is needed here. To test this, try an array of `[1.0.0, 1.potato, 2.0.0]` and you will get a TypeError if I remember correctly (test is to be added)

src/Git/PickLastMinorVersionFromCollection.php
```diff
@@ -19,8 +18,6 @@
      */
     public function forVersions(VersionsCollection $versions) : Version
     {
-        Assert::that($versions->count())->greaterOrEqualThan(1);
-
         $versions->sort(VersionsCollection::SORT_DESC);
```
Ocramius: Relying on internal sorting is indeed risky, and I'm sure @nikolaposa will also make these collections immutable at some point, so it is probably a good idea to cast everything to an `array` via `iterator_to_array()` and operate on that afterwards.

nikolaposa:
@ntzm I've implemented support for parsing `v`-prefixed version strings as you suggested, but `release-` seemed too specific, and it somehow goes beyond the scope of my library. Sorry for letting you down on that one.

As for the issue in question, feel free to open a issue (https://github.com/nikolaposa/version/issues), I would gladly consider fixing it.

src/Git/PickLastMinorVersionFromCollection.php
nikolaposa: @Ocramius TBH I even planned to remove `VersionsCollection` class entirely in a newest 3.0.0 release, thinking it was unnecessary and nobody was using it. :scream: That's why I didn't give enough attention to its design. Feel free to [open an issue](https://github.com/nikolaposa/version/issues) describing fallacies of the class and I'll see what can be done. Of course, if you already have some idea for fixing it, open a PR along the way, I would be glad to accept it.

test/unit/Command/AssertBackwardsCompatibleTest.php
```diff
@@ -280,7 +280,7 @@ public function testExecuteWithDefaultRevisionsNotProvided() : void
     {
         $fromSha       = sha1('fromRevision', false);
         $toSha         = sha1('toRevision', false);
-        $versions      = VersionsCollection::fromArray(['1.0.0', '1.0.1']);
+        $versions      = new VersionsCollection(Version::fromString('1.0.0'), Version::fromString('1.0.1'));
```
nikolaposa: @ntzm I see that this change caused some difficulties, because the ability to create the collection from version strings was very handy I would say now that I'm seeing use cases. Reasoning on my side was making things strict as possible, in order to avoid any ambiguity and simplify validation. So between `string` and `Version` inputs I decided to allow passing `Version` instances only.

I might consider introducing `fromVersionStrings()` and `fromArray()` named constructors for `3.1.0`. Of course, feel free to [suggest features](https://github.com/nikolaposa/version/issues) you might find useful.

src/Git/GetVersionCollectionFromGitRepository.php
ntzm: Seems OK to me: https://3v4l.org/d5RvE

We test invalid at `testFromRepositoryIgnoresInvalidVersions`

test/unit/Command/AssertBackwardsCompatibleTest.php
ntzm: I personally think it's a good change 👍

ntzm:
@nikolaposa I agree with you on the `release-` bit, was just to line up with what composer was looking for

Ocramius:
@nikolaposa I'd probably align to what composer can parse in tags

src/Git/GetVersionCollectionFromGitRepository.php
Ocramius: Hmm, interesting. I guess it crashes only when there are non-numeric keys then :+1:

Ocramius:
About the `->current()` call: do we want to wait for @nikolaposa to add something like `->last()` to the collection instead?

PR #62, opened 2018-05-10
Upgrade to latest BetterReflection version

Ocramius:
This patch:

 * bumps to BetterReflection `master` (no changes in `composer.json` though, just lock file)
 * excludes infection and phpstan, because they are still on PHP-Parser 3.x

Marked as WIP since I don't think excluding this stuff is a good idea, considering that phpstan is being released soon-ish...

### DIFF
--- a/test/unit/Support/ArrayHelpersTest.php
+++ b/test/unit/Support/ArrayHelpersTest.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace RoaveTest\BackwardCompatibility\Support;
 
-use PHPStan\Testing\TestCase;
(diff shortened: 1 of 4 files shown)


test/unit/Support/ArrayHelpersTest.php
```diff
@@ -4,7 +4,7 @@
 
 namespace RoaveTest\BackwardCompatibility\Support;
 
-use PHPStan\Testing\TestCase;
+use PHPUnit\Framework\TestCase;
```
Ocramius: This had slipped in and I noticed it only after removing phpstan

PR #64, opened 2018-05-15
Skip anonymous classes in defined API

### LINKED ISSUE #63: Skip anon classes in "defined API"
Although PR #60 works, @Ocramius mentioned that all the anon classes are detected as "removed classes" and that the anon classes should be removed from the "defined API" up front.

asgrim:
Fixes #63

### DIFF
--- a/src/CompareClasses.php
+++ b/src/CompareClasses.php
@@ -44,9 +44,17 @@ final class CompareClasses implements CompareApi
     ) : Changes {
         $changelog = Changes::empty();
 
-        $definedApiClassNames = array_map(function (ReflectionClass $class) : string {
-            return $class->getName();
-        }, $definedSymbols->getAllClasses());
+        $definedApiClassNames = array_map(
+            function (ReflectionClass $class) : string {
+                return $class->getName();
+            },
+            array_filter(
+                $definedSymbols->getAllClasses(),
+                function (ReflectionClass $class) : bool {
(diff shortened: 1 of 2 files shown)


test/unit/CompareClassesTest.php
```diff
@@ -152,6 +152,22 @@ public function testWillRunTraitComparators() : void
         );
     }
 
+    public function testAnonymousClassesAreFilteredOut() : void
+    {
+        $this->classBasedComparatorWillNotBeCalled();
+        $this->interfaceBasedComparatorWillNotBeCalled();
+        $this->traitBasedComparatorWillNotBeCalled();
+
+        self::assertEqualsIgnoringOrder(
```
Ocramius: Can just be `assertEquals`

src/CompareClasses.php
```diff
@@ -44,9 +44,17 @@ public function __invoke(
...
-            return $class->getName();
-        }, $definedSymbols->getAllClasses());
+        $definedApiClassNames = array_map(
+            function (ReflectionClass $class) : string {
+                return $class->getName();
+            },
+            array_filter(
+                $definedSymbols->getAllClasses(),
+                function (ReflectionClass $class) : bool {
+                    return !$class->isAnonymous();
```
Ocramius: I think a space is needed here

PR #65, opened 2018-05-24
Use phar for infection and phpstan

BackEndTea:
This PR:

* Makes use of phars for infection and phpstan
* Does so by use of a `Makefile`
* Adds an `editorconfig`to make sure the Makefile uses tabs

Another option would be to make use of composer scripts, but i prefer a Makefile.

A downside of this is that `composer update` won't have an effect on these dependencies. Meaning one would have to manually update their version in the makefile.

Removed/changed dependencies:
For infection:
```
  - Removing symfony/yaml (v4.0.8)
  - Removing psr/container (1.0.0)
  - Removing pimple/pimple (v3.2.3)
  - Removing padraic/phar-updater (v1.0.6)
  - Removing padraic/humbug_get_contents (1.1.2)
  - Removing infection/infection (0.8.1)
  - Installing symfony/polyfill-ctype (v1.8.0): Loading from cache
  - Updating symfony/filesystem (v4.0.8 => v4.0.10): Downloading (100%)         
  - Updating symfony/finder (v4.0.8 => v4.0.10): Loading from cache
```

For phpstan:
```
  - Removing phpstan/phpstan (0.9.2)
  - Removing phpstan/phpdoc-parser (0.2)
  - Removing ocramius/package-versions (1.3.0)
  - Removing nette/utils (v2.5.1)
  - Removing nette/robot-loader (v3.0.3)
  - Removing nette/php-generator (v3.0.4)
  - Removing nette/neon (v2.4.2)
  - Removing nette/finder (v2.4.1)
  - Removing nette/di (v2.4.11)
  - Removing nette/bootstrap (v2.4.5)
  - Removing jean85/pretty-package-versions (1.1)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ jobs:
   include:
     - stage: Static analysis (src) with phpstan
       php: 7.2
-      script: vendor/bin/phpstan analyse --level=7 src
+      script: make phpstan
 
     - stage: Test
       php: 7.2
@@ -37,7 +37,7 @@ jobs:
 
     - stage: Mutation Tests
       php: 7.2
-      script: vendor/bin/infection
+      script: make infection
 
     - stage: Verify BC Breaks
       php: 7.2
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,17 @@
+PHONY: infection phpstan
+./infection.phar:
+	wget https://github.com/infection/infection/releases/download/0.8.1/infection.phar
+	chmod a+x ./infection.phar
+
+./infection.phar.pubkey:
+	wget https://github.com/infection/infection/releases/download/0.8.1/infection.phar.pubkey
+
+infection: infection.phar infection.phar.pubkey
+	./infection.phar
+
+./phpstan.phar:
+	wget https://github.com/phpstan/phpstan/releases/download/0.9.2/phpstan.phar
+	chmod a+x ./phpstan.phar
+
+phpstan: phpstan.phar
+	./phpstan.phar analyse --level=7 src
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,2 @@
+[Makefile]
+indent_style=tab
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 vendor
 infection-log.txt
+
+*.phar*
--- a/composer.json
+++ b/composer.json
@@ -19,8 +19,6 @@
     ],
     "require-dev": {
         "doctrine/coding-standard": "^4.0",
-        "infection/infection": "^0.8.1",
-        "phpstan/phpstan": "^0.9.2",
         "phpunit/phpunit": "^7.0",
         "squizlabs/php_codesniffer": "^3.2",
         "roave/security-advisories": "dev-master"
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "80e7f7c2302252715849f4426c57d1f7",
+    "content-hash": "afb7aacc05d1e3d47f81fabebcfb38a4",
     "packages": [
         {
             "name": "beberlei/assert",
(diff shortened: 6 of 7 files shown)


.travis.yml
```diff
@@ -16,7 +16,7 @@ jobs:
   include:
     - stage: Static analysis (src) with phpstan
       php: 7.2
-      script: vendor/bin/phpstan analyse --level=7 src
+      script: make phpstan
```
Ocramius: Please just put the executed code here, no intermediate build tools

.travis.yml
```diff
@@ -37,7 +37,7 @@ jobs:
 
     - stage: Mutation Tests
       php: 7.2
-      script: vendor/bin/infection
+      script: make infection
```
Ocramius: Same: to be reverted to bare bash

Makefile
```diff
@@ -0,0 +1,17 @@
+PHONY: infection phpstan
+./infection.phar:
+	wget https://github.com/infection/infection/releases/download/0.8.1/infection.phar
+	chmod a+x ./infection.phar
+
+./infection.phar.pubkey:
```
Ocramius: Until all phars are completely scoped, using phar distributions is out of question due to autoloaded dependency conflicts

BackEndTea: Infection, in the versions below `0.9.0` does not autoload any code.

BackEndTea:
> Hmm... not sure what this adds, what is the benefit here?

It reduces issues like the ones seen in #62 
@sebastianbergmann did a pretty good job explaining it in this thread: https://twitter.com/s_bergmann/status/999635212723212288

.travis.yml
BackEndTea: Would it be okay to still use `make ./phpstan.phar` to download the phar? or do you want the full wget here as well? That could cause differences in the versions used for local development and the version used on travis.

Ocramius:
That thread is flawed because it doesn't consider static type collisions,
which happen with all the tools currently in use .

.travis.yml
Ocramius: I'd probably rather have a `composer.lock` handle that, heh

asgrim:
> > Hmm... not sure what this adds, what is the benefit here?
> 
> It reduces issues like the ones seen in #62 
> @sebastianbergmann did a pretty good job explaining it in this thread: https://twitter.com/s_bergmann/status/999635212723212288

This is very interesting, and yes it's a good solution, but I have a big "BUT":

> A downside of this is that composer update won't have an effect on these dependencies. Meaning one would have to manually update their version in the makefile.

That's problematic for local development if we have to _separately_ download deps like infection/phpstan... (yes, I do run these tools locally). The collisions seen in #62 are indeed unfortunate, but they don't always happen... and it's only temporary. This seems to suggest instead that we manually install tools locally.

TL;DR - I can see benefits to both sides (thank you for bringing this to our attention!) but my personal vote would be to keep these tools managed by `composer`...

BackEndTea:
> That's problematic for local development if we have to separately download deps like infection/phpstan... 

Currently your standard work flow would be:
```bash
$ composer install
$ vendor/bin/infection
```
Using the makefile it would now be:
```bash
$ composer install
$ make infection
```
I added the `Makefile` to ease the installation of the tools, so you don't have to manually install these tools.

The updating workflow however now does change. Simply running `composer update` will not update these dependencies, that has to be done by hand.

You could also consider using a tool like https://github.com/phar-io/phive for that.

Ocramius:
I'm still firmly set on using composer dependencies. PHAR distributions will only be an acceptable way of sharing dependencies once https://github.com/humbug/php-scoper is working for all projects (that we rely upon) that are being distributed as PHARs.

BackEndTea:
For what its worth: the upcoming version of infection will be fully scoped

Ocramius:
Closing here: we'll have to wait for fully scoped PHARs before being able to move forward, and I'm generally against using PHARs for dependencies (any of them), as the SAT process is a critical/vital part of the setup and of the entire SemVer ecosystem. 

In a library that (hopefully can) proud itself to push the boundaries of BC break detection, I think it is a better idea to give out a signal to consumers that dependency constraints, resolution and version upgrades should be automated via a repository (composer's), so even if the PHARs improve, I'd rather like to have a dependency on a composer package that represents the scoped deliverable, rather than a file fetched from a random location.

PR #66, opened 2018-05-26
fix when no from option is given

bendavies:
`hasOption` => `Returns true if an InputOption object exists by name.`

### DIFF
--- a/src/Command/AssertBackwardsCompatible.php
+++ b/src/Command/AssertBackwardsCompatible.php
@@ -148,7 +148,7 @@ USAGE
         // @todo fix flaky assumption about the path of the source repo...
         $sourceRepo = CheckedOutRepository::fromPath(getcwd());
 
-        $fromRevision = $input->hasOption('from')
+        $fromRevision = $input->getOption('from') !== null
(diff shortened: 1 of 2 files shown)


src/Command/AssertBackwardsCompatible.php
```diff
@@ -148,7 +148,7 @@ public function execute(InputInterface $input, OutputInterface $output) : int
         // @todo fix flaky assumption about the path of the source repo...
         $sourceRepo = CheckedOutRepository::fromPath(getcwd());
 
-        $fromRevision = $input->hasOption('from')
+        $fromRevision = $input->getOption('from') !== null
```
Ocramius: This kind of change requires a test addition

bendavies: Yep sorry, didn't see the command was tested. Will add.

bendavies: test already exists, updated it.

Ocramius:
@bendavies I think #73 fixed this already (by using the correct API in `symfony/console`, which has broken `hasOption()`): closing as `duplicate` here, sorry :-\

PR #71, opened 2018-05-27
#70 look for further `vendor/autoload.php` locations to support in-project-dev-dependency installations

### LINKED ISSUE #70: Broken autoload when in vendor/bin
I installed this on a project that doesn't have a specified bin-dir

```bash
> vendor/bin/roave-backward-compatibility-check 
PHP Fatal error:  Uncaught RuntimeException: Could not find Composer autoload.php in /my/project/vendor/roave/backward-compatibility-check/bin/roave-backward-compatibility-check.php:292
```

Debugging the loop in `roave-backward-compatibility-check.php` it's searching the following locations:

```
/my/project/vendor/roave/backward-compatibility-check/bin/../vendor/autoload.php
/my/project/vendor/roave/backward-compatibility-check/bin/../autoload.php
```

Of course neither contain the project deps because they're all higher up in `/my/project/vendor`

We tackle the similar issue in phpspec by also looking at paths relative to `getcwd()` but it's not that great a solution.

At the least, you can document that `bin-dir` needs to exist _and_ it needs to be one dir below the project root

Ocramius:
Closes #70 

I was trying to find a way to test this via `.travis.yml`, but the problem is that I'd need to add a composer `vcs` repository and then use a `HEAD` reference with some ugly hacks that will break everything, so I'd probably skip that for now.

@ciaranmcnulty can you please review this one? Use `?w=1` :-)

### DIFF
--- a/bin/roave-backward-compatibility-check.php
+++ b/bin/roave-backward-compatibility-check.php
@@ -33,7 +33,7 @@ use function file_exists;
 use function getcwd;
 
 (function () : void {
-    (function () {
+    (function () : void {
         $autoloaderLocations = [
             __DIR__ . '/../vendor/autoload.php',
             __DIR__ . '/../autoload.php',
--- a/bin/roave-backward-compatibility-check.php
+++ b/bin/roave-backward-compatibility-check.php
@@ -30,148 +30,86 @@ use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Console\Output\ConsoleOutput;
 use function file_exists;
+use function getcwd;
 
 (function () : void {
-    foreach ([__DIR__ . '/../vendor/autoload.php', __DIR__ . '/../autoload.php'] as $autoload) {
-        if (! file_exists($autoload)) {
-            continue;
+    (function () {
+        $autoloaderLocations = [
+            __DIR__ . '/../vendor/autoload.php',
+            __DIR__ . '/../autoload.php',
+            __DIR__ . '/../../autoload.php',
+            getcwd() . '/vendor/autoload.php',
+        ];
+
+        foreach ($autoloaderLocations as $autoload) {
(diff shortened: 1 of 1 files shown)


bin/roave-backward-compatibility-check.php
```diff
@@ -30,158 +30,106 @@
...
+            __DIR__ . '/../../autoload.php',
+            getcwd() . '/vendor/autoload.php',
+        ];
+
+        foreach ($autoloaderLocations as $autoload) {
+            if (file_exists($autoload)) {
+                /** @noinspection PhpIncludeInspection */
+                require_once $autoload;
+
+                return;
```
ciaranmcnulty: There's an argument for not breaking when the first one is found, depending what use cases you want to support.

If I build a PHAR or do a `composer global require`, or any other method where the tool has its own deps locally then I need to include both the tool's autoloader _and_ the project's autoloader

ciaranmcnulty: Wait this might not apply in your case if you're not depending on project reps being loaded

Ocramius: > I need to include both the tool's autoloader and the project's autoloader

This can happen when using classes from this project directly, but not with the `bin/roave-backward-compatibility-check.php`, which is closed for customisation.

ciaranmcnulty: Not sure what you mean...

I mean does this tool need to autoload target project's classes?

Ocramius: No, it doesn't, but you can use classes from this tool in your own project

ciaranmcnulty: OK then it's sufficient to find 'our' autoloader 👍

Ocramius: Then :ship:

Lemme finish this :beer: and I'll release

Issue #72, opened 2018-05-27
Git error when running without --from option


ciaranmcnulty:
Steps to reproduce:

1. check out `phpspec/phpspec`
2. add this tool to composer requires and update
3. run `vendor/bin/roave-backward-compatibility-check`

Error:

```
[Symfony\Component\Process\Exception\ProcessFailedException]                     
  The command "'git' 'rev-parse' ''" failed.                                       
                                                                                   
  Exit Code: 128(Invalid exit argument)
```

The input parameter `--from` is optional so its default value is `NULL`. this is explicitly cast to a string `''` in `AssertBackwardsCompatible::parseRevisionFromInput`, but the implementation of `GitParseRevision` then builds a the CLI with an empty string, which breaks as above

ciaranmcnulty:
oh I see the problem, will PR

Ocramius:
Handled in #73

PR #73, opened 2018-05-27
Check that from is actually provided

ciaranmcnulty:
This addresses #72 - in my instance `hasOption` is returning true because its been replaced with its default value of `null`. 

I'm mildly surprised that this hasn't happened to anyone pre-release though - are you always specifying a revision? It's possible that this is a symfony version thing.

I stuck to the existing paradigm for the tests but there's a _lot_ of stubbing going on - it might be clearer to use a concrete ArgvInput (which would also have avoided the issue of stubbing enshrining an incorrect understanding of the API)

### DIFF
--- a/src/Command/AssertBackwardsCompatible.php
+++ b/src/Command/AssertBackwardsCompatible.php
@@ -148,7 +148,7 @@ USAGE
         // @todo fix flaky assumption about the path of the source repo...
         $sourceRepo = CheckedOutRepository::fromPath(getcwd());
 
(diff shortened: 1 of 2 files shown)


Ocramius:
Hmm, no, this patch completely killed `--from` support, as a precise value is being expected here. 

I'll probably have to revert it completely and apply #66 instead.

ciaranmcnulty:
Sound fair - but write some end-to-end tests

PR #76, opened 2018-05-28
#66 #72 #73 fixed missing `--from` handling, minor version detection and scenario with no detected versions at all

### LINKED ISSUE #72: Git error when running without --from option

Ocramius:
Fixes #66 
Fixes #72 
Fixes #73


Ocramius:
@bendavies I integrated your commits.

@ciaranmcnulty any suggestion for scenarios that I may not have covered yet?

test/e2e/Command/AssertBackwardsCompatibleTest.php
```diff
@@ -0,0 +1,301 @@
...
+    {
+        (new Process(
+            [
+                'git',
+                'checkout',
+                $this->versions[1],
+            ],
+            $this->sourcesRepository
+        ))->mustRun();
+        (new Process(
```
Ocramius: These operations should likely be isolated in a `private` method

test/e2e/Command/AssertBackwardsCompatibleTest.php
```diff
@@ -0,0 +1,301 @@
...
+
+        self::assertContains('Detected last minor version: 1.2.3', $errorOutput);
+        self::assertStringEndsWith(
+            <<<'EXPECTED'
+[BC] CHANGED: The parameter $a of TestArtifact\TheClass#method() changed from TestArtifact\B to a non-contravariant TestArtifact\C
+1 backwards-incompatible changes detected
+
+EXPECTED
+            ,
+            $errorOutput // @TODO this looks like a symfony bug - we shouldn't check STDERR, but STDOUT
```
Ocramius: This is a problem. For some reason, everything is going to STDERR in my tests (tried manually via `bash`, so the bug is with `symfony/console` or with how we use it)

test/e2e/Command/AssertBackwardsCompatibleTest.php
```diff
@@ -0,0 +1,301 @@
...
+    "repositories": [
+        {
+            "packagist.org": false
+        }
+    ]
+}
+
+JSON;
+
+    private const CLASS_V1 = <<<'PHP'
```
Ocramius: These constants should be moved to an array

test/e2e/Command/AssertBackwardsCompatibleTest.php
bendavies: https://github.com/symfony/symfony/pull/13730 ?

Ocramius: Looked it up as well - unsure if that's the actual bug. I'll probably need to dump the streams and see if both point to STDERR by accident, because the output sure only is on STDERR.

asgrim: Umm, we are manually splitting output between stderr/stdout ourselves.

If you specify `--format=markdown`, `stdout` is formatted markdown: `stderr` contains the human-readable output, so this is the expectation. Not a symfony bug.

See these for `stderr` being used:

https://github.com/Roave/BackwardCompatibilityCheck/blob/9e95a606a84f398212b17287aa7801d39c72e460/src/Command/AssertBackwardsCompatible.php#L146

https://github.com/Roave/BackwardCompatibilityCheck/blob/9e95a606a84f398212b17287aa7801d39c72e460/src/Command/AssertBackwardsCompatible.php#L151-L153

https://github.com/Roave/BackwardCompatibilityCheck/blob/9e95a606a84f398212b17287aa7801d39c72e460/src/Command/AssertBackwardsCompatible.php#L185

And then, markdown formatting if required:

https://github.com/Roave/BackwardCompatibilityCheck/blob/9e95a606a84f398212b17287aa7801d39c72e460/src/Command/AssertBackwardsCompatible.php#L190-L192

This was done specifically so you can pipe output to a file - see PR #35 , specifically https://github.com/Roave/BackwardCompatibilityCheck/pull/35#discussion_r181588762

Ocramius: Yes, but in the test (also manual ones), redirecting output doesn't seem to have any effect

Ocramius: Moved to https://github.com/Roave/BackwardCompatibilityCheck/issues/79

Ocramius:
:ship:

PR #78, opened 2018-05-30
Throw exception if repository does not contain any tagged version

OndraM:
Hi,
I spend a while debugging failing Travis build which uses roave-backward-compatibility-check, because I was getting this error:

```
PHP Fatal error:  Uncaught TypeError: Return value of Roave\BackwardCompatibility\Git\PickLastMinorVersionFromCollection::forVersions() must be an instance of Version\Version, boolean returned in /.../src/Git/PickLastMinorVersionFromCollection.php:43
Stack trace:
#0 /.../src/Command/AssertBackwardsCompatible.php(230): Roave\BackwardCompatibility\Git\PickLastMinorVersionFromCollection->forVersions(Object(Version\VersionsCollection))
#1 /.../src/Command/AssertBackwardsCompatible.php(153): Roave\BackwardCompatibility\Command\AssertBackwardsCompatible->determineFromRevisionFromRepository(Object(Roave\BackwardCompatibility\Git\CheckedOutRepository), Object(Symfony\Component\Console\Output\StreamOutput))
#2 /.../vendor/symfony/console/Command/Command.php(251): Roave\BackwardCompatibility\Command\AssertBackwardsCompatible->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symf in /srv/www/test/BackwardCompatibilityCheck/src/Git/PickLastMinorVersionFromCollection.php on line 43
```

which was really confusing. After some debugging I finally realized this was caused by running the check on a fork, which does not have any tags pushed.

This is why I suggest something like this, to make it clear what was the reason of failure in case repository does not contain any tagged versions.

What do you think?

### DIFF
--- a/src/Git/PickLastMinorVersionFromCollection.php
+++ b/src/Git/PickLastMinorVersionFromCollection.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Roave\BackwardCompatibility\Git;
 
+use Assert\Assert;
 use Symfony\Component\Process\Exception\LogicException;
 use Symfony\Component\Process\Exception\RuntimeException;
 use Version\Version;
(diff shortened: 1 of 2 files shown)


Ocramius:
I think this was fixed in https://github.com/Roave/BackwardCompatibilityCheck/pull/76 as well, can you check it?

Ocramius:
@OndraM if you can review that PR carefully, that would help me further :100:

PR #84, opened 2018-06-21
Update README.md

alberto-bottarini:
add php version requirement


Ocramius:
Heya!

Sorry, but the project can use any PHP version understood by PHP-Parser.

The only thing that is required is PHP 7.2 in your build process (which will run the diffing): that is already part of `composer.json`, and therefore redundant in the documentation.

PR #88, opened 2018-07-02
Turn on checks for mixed types

muglug:
This PR attempts to eradicate any `mixed` types in this codebase. It does this with the help of Psalm's `totallyTyped="true"` config.

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -5,7 +5,7 @@
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
     name="Example Psalm config with recommended defaults"
     useDocblockTypes="true"
-    totallyTyped="false"
+    totallyTyped="true"
     allowPhpStormGenerics="true"
 >
     <projectFiles>
@@ -13,6 +13,27 @@
     </projectFiles>
 
     <issueHandlers>
-        <PropertyNotSetInConstructor errorLevel="suppress"/>
+        <PropertyNotSetInConstructor>
+            <errorLevel type="suppress">
+                <file name="src/Git/CheckedOutRepository.php" />
+                <file name="src/Git/Revision.php" />
+                <file name="src/Changes.php" />
+            </errorLevel>
+        </PropertyNotSetInConstructor>
+
+        <MixedAssignment>
+            <errorLevel type="suppress">
+                <file name="src/DetectChanges/BCBreak/ClassConstantBased/ClassConstantValueChanged.php" />
+                <file name="src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php" />
+                <file name="src/DetectChanges/BCBreak/PropertyBased/PropertyDefaultValueChanged.php" />
+                <file name="src/LocateDependencies/LocateDependenciesViaComposer.php" />
+            </errorLevel>
+        </MixedAssignment>
+
+        <MixedArgument>
+            <errorLevel type="suppress">
+                <file name="src/LocateDependencies/LocateDependenciesViaComposer.php" />
+            </errorLevel>
+        </MixedArgument>
     </issueHandlers>
 </psalm>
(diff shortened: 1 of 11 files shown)


psalm.xml
```diff
@@ -5,14 +5,35 @@
...
+    totallyTyped="true"
     allowPhpStormGenerics="true"
 >
     <projectFiles>
         <directory name="src"/>
     </projectFiles>
 
     <issueHandlers>
-        <PropertyNotSetInConstructor errorLevel="suppress"/>
+        <PropertyNotSetInConstructor>
```
Ocramius: Named constructors will be used in every value type in the library (and tbh in all the new code I'm writing these days): does this check make any sense at all? Are named constructors something that the library could work with? Effectively, the rule should be something like "complete this object before it is passed to anything else", rather than "complete this object in the constructor".

psalm.xml
```diff
@@ -5,14 +5,35 @@
...
+            <errorLevel type="suppress">
+                <file name="src/Git/CheckedOutRepository.php" />
+                <file name="src/Git/Revision.php" />
+                <file name="src/Changes.php" />
+            </errorLevel>
+        </PropertyNotSetInConstructor>
+
+        <MixedAssignment>
```
Ocramius: Constants and default values can be a recursive structure `data ConstValue = int | int | string | float | bool | null | array ConstValue`, not really `mixed`

src/DetectChanges/BCBreak/FunctionBased/ParameterByReferenceChanged.php
```diff
@@ -30,9 +30,9 @@ public function __construct()
 
     public function __invoke(ReflectionFunctionAbstract $fromFunction, ReflectionFunctionAbstract $toFunction) : Changes
     {
-        /** @var ReflectionParameter[] $fromParameters */
+        /** @var array<int, ReflectionParameter> $fromParameters */
```
Ocramius: As discussed elsewhere (I think it was BetterReflection), this currently clashes with JetBrains tooling, and the `array_values()` already implies the `int[]` keys

psalm.xml
muglug: Yeah, I'll get rid of this if it's not how you write code.

> Are named constructors something that the library could work with?
> Effectively, the rule should be something like "complete this object before it is passed to anything else", rather than "complete this object in the constructor".

I think it could be fairly easily supported as long as it only applied to classes with

```
private function __construct {}
```

There'd be some other error (`PropertyNotSetAfterInitialization` or similar) in that circumstance.

psalm.xml
muglug: Yeah but `mixed` basically means "I don't know what this is", appropriate for this use case. I guess I could add a `mixed-const` type, with rules around that, but `mixed` is far simpler. If I added `mixed-const` I'd also want to add `mixed-array-key` to represent array types (currently represented by `mixed` even though they can only be `string` or `int`).

src/DetectChanges/BCBreak/FunctionBased/ParameterByReferenceChanged.php
muglug: > currently clashes with JetBrains tooling

uggghhh - ok, will change

psalm.xml
asgrim: Yeh that would make sense :+1:

src/DetectChanges/BCBreak/FunctionBased/ParameterByReferenceChanged.php
Majkl578: Yes, this is [7 years old issue](https://youtrack.jetbrains.com/issue/WI-6558) that is still not addressed.
Right now Jetbrains guys are [claiming to be waiting for PSR-5](https://twitter.com/neuro159/status/1029335046325514240), which doesn't really fix anything (no proper support for generics - only collections, no intersections, ...) and doesn't even evolve. Based on that, I don't expect any progress or useful outcome in the near future.

Ocramius: Looks like at least PSR-5 is going that direction (for arrays)

Ocramius:
This is redundant with improvements that already landed in `master` last year :+1:

PR #91, opened 2018-08-20
fix: detect the auto-loader of the project you're installed in

dkarlovi:
Relative to `roave/backward-compatibility-check/bin/`

Visible if installed using `composer-bin`.

### DIFF
--- a/bin/roave-backward-compatibility-check.php
+++ b/bin/roave-backward-compatibility-check.php
@@ -38,6 +38,7 @@ use function getcwd;
             __DIR__ . '/../vendor/autoload.php',
             __DIR__ . '/../autoload.php',
             __DIR__ . '/../../autoload.php',
+            __DIR__ . '/../../../autoload.php',
             getcwd() . '/vendor/autoload.php',
         ];
 


dkarlovi:
```
$ mkdir rcc && cd rcc
$ composer require roave/backward-compatibility-check
$ cd vendor/ # if here, it works as if by chance
$ roave/backward-compatibility-check/bin/roave-backward-compatibility-check 
PHP Fatal error:  Uncaught RuntimeException: Could not find Composer autoload.php 
$ ./bin/roave-backward-compatibility-check
PHP Fatal error:  Uncaught RuntimeException: Could not find Composer autoload.php 
```

etc.

asgrim:
You shouldn't need to `cd vendor` just run `vendor/bin/roave-backward-compatibility-check` in the root of your project after requiring it (side note: probably best to use `--dev`)

dkarlovi:
If you're using [Composer bin plugin](https://github.com/bamarni/composer-bin-plugin) (directly or indirectly, like in referenced issue), it breaks.

dkarlovi:
@asgrim 
> You shouldn't need to cd vendor just run vendor/bin/roave-backward-compatibility-check in the root of your project after requiring it

This was just a reproducer, I'm not using it like that at all, it's installed in a Docker image used as a portable QA/tooling environment.

dkarlovi:
@asgrim this one is actually faulty since it loads an autoloader based on where in the filesystem you are. Roll a dice, load an autoloader! :)

https://github.com/Roave/BackwardCompatibilityCheck/blob/c38d31e61e5d252b361ab50869b69c26edc69872/bin/roave-backward-compatibility-check.php#L41

dkarlovi:
Fair, but in that case you should probably remove all the auto-loader paths except the one you support as a use case.

You can even remove the whole autoloader heuristic, come to think about it. :+1:

Ocramius:
@dkarlovi currently removing all paths that are not in the pre-determined locations

Ocramius:
Note: manually merged this patch anyway, but the path is a replacement for existing ones, not an additional one :+1:

Ocramius:
Included in #94

PR #92, opened 2018-08-20
#85 fail-safe wrapper around BC compliance checks that access risky/throw-happy reflection API

### LINKED ISSUE #85: Add `Throwable`-safe wrappers around all `Roave\BackwardCompatibility\DetectChanges\BCBreak\*` classes that may crash during reflection operations
The idea is that a BC break should be detected (following the mantra "when in doubt: something is wrong"), and the exception message collected and shown in the report, but without crashing further execution.

Ocramius:
Fixes #85

### DIFF
--- a/src/Change.php
+++ b/src/Change.php
@@ -5,9 +5,9 @@ declare(strict_types=1);
 namespace Roave\BackwardCompatibility;
 
 use Assert\Assert;
+use Throwable;
 use function sprintf;
 use function strtoupper;
-use Throwable;
 
 /**
  * @todo this class probably needs subclassing or being turned into an interface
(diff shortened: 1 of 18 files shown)


src/Change.php
```diff
@@ -52,6 +54,12 @@ public static function removed(string $description, bool $isBcBreak) : self
         return new self(self::REMOVED, $description, $isBcBreak);
     }
 
+    public static function skippedDueToFailure(Throwable $failure) : self
+    {
+        // @TODO Note: we may consider importing the full exception for better printing later on
```
Ocramius: Moved to https://github.com/Roave/BackwardCompatibilityCheck/issues/93

Ocramius:
Quoting @asgrim:

 > @ocramius I see good shit happening atm on BCC, but I'm behind on like everything atm
 > I'll catch up but if it's urgent please just merge shit
 > eventuall

So lots of shit, which means :ship:, I suppose :+1:

PR #94, opened 2018-08-21
Remove CWD usage from autoload discovery paths, limit supported installations to `composer require`, `composer create-project` or `git clone` + `composer install`

Ocramius:
Includes #91


Ocramius:
Note: BC check build failed because this branch doesn't have all the features that `master` has. :ship:

PR #95, opened 2018-08-21
#69 re-introduced PHPStan static analysis

Ocramius:
Fixes #69
Provides static analysis checks also for `test/`

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,4 @@
+parameters:
+    level: 7
+    paths:
+        - %currentWorkingDirectory%/src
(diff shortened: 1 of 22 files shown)


phpstan.neon
```diff
@@ -0,0 +1,8 @@
+parameters:
```
Majkl578: Why not phpstan.neon.dist?

Ocramius: Missed it, making a new PR

Ocramius: https://github.com/Roave/BackwardCompatibilityCheck/pull/99

PR #97, opened 2018-08-21
Add --sources-from and --sources-to options, deprecate sources-path argument

Majkl578:
From @Ocramius on Slack:

> I think the path should be a parameter rather than an argument

This PR does exactly that. Separate option for --from and --to since it's not rare to see libraries change i.e. lib/ -> src/ between majors.

```
bin/roave-backward-compatibility-check r:a --from 1.2.3 --to master --sources-from lib --sources-to src
```

Advice on testing this? DirectoryReflectorFactory is final without interface. AssertBackwardsCompatibleTest e2e is built around src/ only.

### DIFF
--- a/src/Command/AssertBackwardsCompatible.php
+++ b/src/Command/AssertBackwardsCompatible.php
@@ -100,6 +100,20 @@ final class AssertBackwardsCompatible extends Command
                 'HEAD'
             )
             ->addOption(
+                'sources-from',
+                null,
+                InputOption::VALUE_REQUIRED,
+                'Path to the sources in --from revision, relative to the repository root',
+                'src'
+            )
+            ->addOption(
+                'sources-to',
+                null,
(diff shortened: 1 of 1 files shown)


Ocramius:
As discussed in chat, we should explore auto-discovery via `composer.json` `autoload` before jumping ahead and adding more flags.

Ocramius:
Relevant: https://github.com/Roave/BetterReflection/issues/442

Ocramius:
Possibly superseded by #102

Ocramius:
Note: implemented in #102, but likely still relevant if the `composer.json` is not at the root of the project. Maybe interesting for you for 2.1.0, @staabm?

Ocramius:
This is implicitly handled in #120, which uses `composer.json` to locate sources across versions.

PR #108, opened 2018-09-02
Moved all BC compliance and `Changes` instance to internally use generators for better "on the fly" results

Ocramius:


src/Changes.php
```diff
@@ -33,6 +37,40 @@ public static function empty() : self
         return $empty;
     }
 
+    /** @param Change[] $changes */
+    public static function fromIterator(iterable $changes) : self
+    {
+        $instance = new self();
+
+        $instance->changes   = [];
+        $instance->generator = (function () use ($changes) : Generator {
```
Majkl578: I don't know what this is, but if some dark magic, then it is confusing and just looks like a typo, so -1. :/

src/Changes.php
```diff
@@ -33,6 +37,40 @@ public static function empty() : self
...
+            foreach ($changes as $change) {
+                yield $change;
```
Majkl578: Why not just `yield from $changes`?

src/Changes.php
Ocramius: I just smashed it together while on the plane - the idea is to make the iteration repeatable.

src/Changes.php
Ocramius: :+1:

src/Changes.php
Majkl578: I was talking about ``ffunction` here:

But it seems to be a rendering bug, lol.

Ocramius: It is a rendering bug, yep

Ocramius: This has been massively simplified now :+1:

Ocramius:
There is still a massive issue with this patch: all exceptions are now un-caugh, as they are being delayed until when the actual looping around the iterator is being performed (which happens while rendering)

Ocramius:
Verified further, and the problem is not trivial to solve: would need to wrap every unit-alike BC checker into a `Skip*BasedErrors` instance, and ignore all the generator-alike checkers.

Feasible, just extremely tedious and verbose.

Ocramius:
If anybody feels like willing to review this: [I'm sorry](https://github.com/Roave/BackwardCompatibilityCheck/pull/108/commits/e91f59dc1dc574efff711f5bea3226680f3b2dc1).

Ocramius:
The reported BC break is not really relevant:

```
[BC] CHANGED: The return type of Roave\BackwardCompatibility\Changes#getIterator() changed from ArrayIterator to the non-covariant iterable
```

PR #116, opened 2019-02-03
Added docks how to use Github actions and docker

Nyholm:
I really like this tool. With some inspiration from PHPStan, I created this repo: https://github.com/Nyholm/roave-bc-check-docker

It builds docker images and support for Github actions. 

This PR contains installation instructions that might be easier to use... however, I also understand if you only want to have a link saying: "If you use docker or github actions, go here ->"

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,14 @@ of a PHP library.
 composer require --dev roave/backward-compatibility-check
 ```
 
+### Install with Docker
+
+You can also use Docker to run `roave-backward-compatibility-check`: 
+
+```bash
+docker run --rm -v `pwd`:/app nyholm/roave-bc-check
+```
+
 ## Usage
 
 ### Adding to a continuous integration pipeline
@@ -34,6 +42,24 @@ the build to fail.
 
 *NOTE:* detecting the base version only works if you have git tags in
 the SemVer-compliant `x.y.z` format, such as `1.2.3`.
+
+#### Github action
+
+You can use it as a Github Action like this:
+
+_.github/main.workflow_
+```
+workflow "Main" {
+  on = "push"
+  resolves = ["Roave BC Check"]
+}
+
+action "Roave BC Check" {
+  uses = "docker://nyholm/roave-bc-check-ga"
+  secrets = ["GITHUB_TOKEN"]
+  args = ""
+}
+```
 
 ### Running manually
 


Ocramius:
The patch looks okay, but do you have automation to keep these images fresh/sync'd with tags?

Ocramius:
Any way to also trigger the symc automatically?

Nyholm:
> Any way to also trigger the symc automatically?

If I can trigger the sync for major version automatically? I dont think so. As you can see in https://github.com/Nyholm/roave-bc-check-docker I need different docker file for each major version. I do also need to set up trigger rules on docker hub. 

You **could**  trigger it automatically by either
A) Use a private CI that build and pushes all major versions automatically. (needs my docker API key)
B) Move all the docker files from https://github.com/Nyholm/roave-bc-check-docker to this repo and configure docker hub to build the tags pushed to this repo.

Ocramius:
Note that I'm asking these questions to reduce downstream maintenance. I don't want to land in the scenario (in which I personally already am, for other projects) where somebody needs to run around and do work that can be automated.

 > Move all the docker files from https://github.com/Nyholm/roave-bc-check-docker to this repo and configure docker hub to build the tags pushed to this repo.

Does docker not have scoped API keys? Also, would you be OK with being added as a contributor to the repo, if we moved the scope to something specifically created for this repo??

dkarlovi:
@Nyholm Can't this repo setup a webhook to ping your Dockerhub?

Ocramius:
Yup, that's something I can set up here :+1:

Nyholm:
It is currently built every day. A webhook would sure help docker hub to spend too much resources. But a part from that it does not bring any value

Ocramius:
If it's already built every day, that's fine, no need to overdo it :)

dkarlovi:
The only difference I see is it would be available in sync with tagging here and not a day later.

Ocramius:
@dkarlovi meh, it's already good enough that I get to tag things every few months :P

README.md
```diff
@@ -35,6 +43,24 @@ the build to fail.
...
+_.github/main.workflow_
+```
+workflow "Main" {
+  on = "push"
+  resolves = ["Roave BC Check"]
+}
+
+action "Roave BC Check" {
+  uses = "docker://nyholm/roave-bc-check-ga"
+  secrets = ["GITHUB_TOKEN"]
```
OskarStark: I think the secret is not needed. I did this for my github actions too, I didn’t write or use any github api endpoints in the image

PR #120, opened 2019-04-14
Usage of roave/better-reflection composer source-locators for major performance improvements

Ocramius:
Requires roave/better-reflection#461

~~Note: still fails to work with `symfony/*` classes, which seem to have weird autoloading setup.~~ (fixed - this is because some of the analysed libraries had `symfony/console` in `require-dev` instead of `require` - behaviour is expected)


Ocramius:
Note: couldn't upgrade @vimeo/psalm due to:

```
ERROR: ImplementedReturnTypeMismatch - src/Changes.php:81:37 - The return type 'Traversable<mixed, mixed>' for IteratorAggregate::getIterator is different to the implemented return type for Roave\BackwardCompatibility\Changes::getiterator 'iterable<array-key, Roave\BackwardCompatibility\Change>'
    public function getIterator() : iterable
```

Possibly something to be fixed in stubs (upstream)

composer.json
```diff
@@ -4,12 +4,13 @@
...
+        "beberlei/assert": "^3.2.0",
+        "composer/composer": "^1.8.5",
         "nikolaposa/version": "^3.1.0",
-        "roave/better-reflection": "^3.0",
-        "symfony/console": "^4.0",
-        "symfony/process": "^4.0"
+        "roave/better-reflection": "^3.4.0",
+        "symfony/console": "^4.2.7",
+        "symfony/process": "^4.2.7",
+        "thecodingmachine/safe": "^0.1.15"
```
Ocramius: Yes, this is new, but for my own sanity I need to stop relying on the PHP internal functions when writing code.

test/e2e/Command/AssertBackwardsCompatibleTest.php
```diff
@@ -199,11 +199,14 @@ public function testWillRunSuccessfullyOnNoBcBreaks() : void
...
                 '--to=' . $this->versions[3],
+                '-vvv',
             ],
             $this->sourcesRepository
         );
 
-        self::assertSame(0, $check->run());
+        $check->run();
+
+//        self::assertSame(0, $check->run());
```
asgrim: Does this need to go back in?

src/Command/AssertBackwardsCompatible.php
```diff
@@ -149,7 +151,11 @@ public function execute(InputInterface $input, OutputInterface $output) : int
             ? $this->parseRevisionFromInput($input, $sourceRepo)
             : $this->determineFromRevisionFromRepository($sourceRepo, $stdErr);
 
-        $toRevision = $this->parseRevision->fromStringForRepository($input->getOption('to'), $sourceRepo);
+        $to = $input->getOption('to');
+
+        assert(is_string($to));
```
asgrim: Can we rely on the end user's environment having `assert` enabled, or should we use `beberlei/assert`?

src/Command/AssertBackwardsCompatible.php
```diff
@@ -175,7 +181,8 @@ public function execute(InputInterface $input, OutputInterface $output) : int
             (new SymfonyConsoleTextFormatter($stdErr))->write($changes);
 
             $outputFormats = $input->getOption('format') ?: [];
-            Assert::that($outputFormats)->isArray();
+
+            assert(is_array($outputFormats));
```
asgrim: Same about `assert`...

src/Command/AssertBackwardsCompatible.php
```diff
@@ -208,7 +215,7 @@ private function parseRevisionFromInput(InputInterface $input, CheckedOutReposit
     {
         $from = $input->getOption('from');
 
-        Assert::that($from)->string();
+        assert(is_string($from));
```
asgrim: Same about `assert`...

test/e2e/Command/AssertBackwardsCompatibleTest.php
Ocramius: Yep, this is a mistake 👍

src/Command/AssertBackwardsCompatible.php
Ocramius: I went with `assert()` for psalm compatibility here: the chance of a type mismatch (due to misconfiguration) are extremely low, so a simple `assert()` would work here.

This is a regression caused by symfony widening the type of `getOption` to `string|int|float|null|array` in newer releases.

src/Command/AssertBackwardsCompatible.php
Ocramius: Similar to the above: will keep using `assert()` due to how the command is configured. If this keeps spawning everywhere, we'll need a psalm plugin.

src/Command/AssertBackwardsCompatible.php
Ocramius: Same as above

Ocramius:
As discussed with @asgrim, I'll need to come up with a plugin similar to https://github.com/phpstan/phpstan-beberlei-assert for @vimeo/psalm: that's not affecting the 3 usages of `assert()` introduced here though.

src/Command/AssertBackwardsCompatible.php
stof: note that Symfony only fixed their phpdoc to account for the different cases depending on the config of the option.

Ocramius: Well aware: this is just documentation getting closer to reality :+1:

composer.json
```diff
@@ -4,29 +4,35 @@
     "require": {
         "php": "^7.2",
         "ext-json": "*",
-        "beberlei/assert": "^2.9",
-        "composer/composer": "^1.6",
+        "beberlei/assert": "^3.2.0",
+        "composer/composer": "^1.8.5",
```
staabm: any particular reason why you put this bar that high?

e.g. on travis we only have 1.8.0 and need a extra step to update to ^1.8.5

stof: @staabm this does not affect the version of composer you have installed on your machine. This is a requirement for installing composer itself as a dependency of this package (as a library)

Ocramius: Besides that, I generally always bump all dependencies to their latest stable, unless there is a strong reason not to do so. This already solves the grand majority of issues for most users that may otherwise get an outdated dependency set (and removes trouble around "please upgrade before reporting an issue")

PR #122, opened 2019-04-23
Chore: dependency upgrades, mutation testing, strict type checking in the test suite

Ocramius:
Includes a BC Break: removed some deprecated code (will pop up in CI report)


composer.json
```diff
@@ -24,16 +24,17 @@
         }
     ],
     "require-dev": {
-        "doctrine/coding-standard": "^5.0.1",
-        "infection/infection": "^0.9.3",
+        "doctrine/coding-standard": "^6.0.0",
+        "infection/infection": "^0.10.6",
```
Ocramius: Note: newer `infection/infection` versions fail with bizarre issues - didn't go there (for now)

Ocramius:
Build failure (expected):

```
[BC] REMOVED: Class Roave\BackwardCompatibility\SourceLocator\StubClassSourceLocator has been deleted
```

src/Change.php
```diff
@@ -33,7 +29,6 @@ final class Change
 
     private function __construct(string $modificationType, string $description, bool $isBcBreak)
     {
-        Assert::that($modificationType)->inArray(self::$validModificationTypes);
```
asgrim: Why removing this validation? Doesn't seem to be any alternative type safety being added instead..?

src/Command/AssertBackwardsCompatible.php
```diff
@@ -141,11 +141,7 @@ public function execute(InputInterface $input, OutputInterface $output) : int
         $stdErr = $output->getErrorOutput();
 
         // @todo fix flaky assumption about the path of the source repo...
-        $cwd = getcwd();
-
-        Assert::that($cwd)->string();
-
-        $sourceRepo = CheckedOutRepository::fromPath($cwd);
+        $sourceRepo = CheckedOutRepository::fromPath(getcwd());
```
asgrim: `getcwd()` may return `false` on failure

src/LocateDependencies/LocateDependenciesViaComposer.php
```diff
@@ -34,13 +37,11 @@ public function __construct(
...
     public function __invoke(string $installationPath) : SourceLocator
     {
-        Assert::that($installationPath)->directory();
         Assert::that($installationPath . '/composer.json')->file();
 
         $this->runInDirectory(function () use ($installationPath) : void {
             $installer = ($this->makeComposerInstaller)($installationPath);
 
-            Assert::that($installer)->isInstanceOf(Installer::class);
             assert($installer instanceof Installer);
```
asgrim: We can't really assume that `assert` will work on all platforms since it is configurable out of our control, so the `beberlei/assert` should remain, since the `makeComposerInstaller` property is just `callable` and has no defined type as the return value

src/LocateDependencies/LocateDependenciesViaComposer.php
```diff
@@ -62,8 +63,6 @@ private function runInDirectory(callable $callable, string $directoryOfExecution
     {
         $originalDirectory = getcwd();
 
-        Assert::that($originalDirectory)->string();
```
asgrim: `getcwd` may return `false`, so assertion is needed

src/SourceLocator/StaticClassMapSourceLocator.php
```diff
@@ -29,7 +32,7 @@ public function __construct(
         $realPaths = array_map('realpath', $classMap);
 
         Assert::that($classMap)->all()->file();
-        Assert::that(array_keys($classMap))->all()->string()->notEmpty();
```
asgrim: `notEmpty` works on non-`string` values so we should keep the `string()` assertion

src/SourceLocator/StaticClassMapSourceLocator.php
```diff
@@ -49,10 +52,6 @@ protected function createLocatedSource(Identifier $identifier) : ?LocatedSource
             return null;
         }
 
-        $fileContents = file_get_contents($classFile);
-
-        Assert::that($fileContents)->string();
```
asgrim: `file_get_contents` may return `false`, so assertion should remain

src/Change.php
Ocramius: All of the API is `private`, so this assertion was not adding any further safety: no public API allows for invalid states

src/Command/AssertBackwardsCompatible.php
Ocramius: Now using `Safe\getcwd`

src/LocateDependencies/LocateDependenciesViaComposer.php
Ocramius: It is a `callable () : Installer` - that wouldn't type-check anymore, hence why I removed the assertion. Can re-introduce it, but can't write a test that triggers this assertion.

src/LocateDependencies/LocateDependenciesViaComposer.php
Ocramius: Similar to above, now using `Safe\getcwd`

src/SourceLocator/StaticClassMapSourceLocator.php
Ocramius: All values here are guaranteed to be strings by type checks: it is not possible to trigger this assertion via a test anymore either

src/SourceLocator/StaticClassMapSourceLocator.php
Ocramius: Similar to the above, now using `Safe\file_get_contents`

src/LocateDependencies/LocateDependenciesViaComposer.php
asgrim: Surely doing `new LocateDependenenciesViaComposer(function () {}, $astLocator)` is enough to trigger this scenario...?

Ocramius: That doesn't type-check anymore, thanks to Psalm

asgrim: Doesn't mean it can't be done at runtime, by any consumer of the library....

src/SourceLocator/StaticClassMapSourceLocator.php
Ocramius: To be clear, we had a test scenario for this, but refining the type declaration led me to remove the test too (and subsequently the assertion, as per mutation tests)

src/Change.php
asgrim: OK gotcha

src/LocateDependencies/LocateDependenciesViaComposer.php
Ocramius: I'd rather say that consumers should respect the contract here: we've been doing this sort of assumption for ages before PHP 7, so I think we should keep doing it, especially since we now can provide explicit tooling for this

src/SourceLocator/StaticClassMapSourceLocator.php
asgrim: This *IS* the type check: `@param` is not sufficient, nor is the `array` type declaration to be able to trust that the keys are all `string`s

src/LocateDependencies/LocateDependenciesViaComposer.php
Ocramius: (and yes, generic types in the engine would be better)

asgrim: We put strict types and assertions every where to avoid consumers mis-using the library and basically making the API foolproof. By removing stuff like this, you are trusting that all consumers understand what `callable () : Installer` actually means (I for example, haven't seen the syntax before, it's new and unclear, so I just assume "oh this needs a callable"). Whilst the annotations are good for static analysis, it's never enforced at runtime, so we need assertions still in place, and NOT using `assert` (because it is effectively useless due to it relying on consumer configuration).

src/SourceLocator/StaticClassMapSourceLocator.php
Ocramius: The type check happens statically, not at runtime.

As of current type declarations, it is impossible to write a test that type-checks (before runtime) and then triggers this assertion (mutation test requirement)

asgrim: It's not impossible:

```php
new StaticClassMapSourceLocator([1 => 'foo'], $astLocator)
```

src/LocateDependencies/LocateDependenciesViaComposer.php
Ocramius: > We put strict types and assertions every where to avoid consumers mis-using the library and basically making the API foolproof. 

Consumers are still very much free to call our API without `declare(strict_types=1);` in their code, so the fool-proofing is mostly in our internal method calls.

> By removing stuff like this, you are trusting that all consumers understand what `callable () : Installer` actually means (I for example, haven't seen the syntax before, it's new and unclear, so I just assume "oh this needs a callable").

Absolutely clear that most people will not have seen this before, bit I think we (@Roave) are trend-setters, and we should push the ecosystem forward on this stuff. I did start pushing this into `doctrine/collections` too, hoping that the PHP ecosystem picks it up 👍

> Whilst the annotations are good for static analysis, it's never enforced at runtime

This is correct and intentional

> so we need assertions still in place

This is where we diverge: it is impossible to have a program that type-checks and triggers type assertions. In order to do that, I would have to write tests that intentionally do not respect the type declarations either.

Besides this technical limitation (which I believe to be correct, not an accident, but rather a happy combo of tooling discovering controversial definitions), I think that type checks are to be performed before runtime. Yes, the PHP engine is "special", because type checks are also applied at runtime, but IMO that shouldn't matter, if a program can be proven to be wrong before a single line of code is executed.



> and NOT using assert (because it is effectively useless due to it relying on consumer configuration).

Fully agree on this point: `assert()` is just bad, and as discussed before, I'll put effort into adding plugins for Psalm here.

src/SourceLocator/StaticClassMapSourceLocator.php
Ocramius: That's not respecting the declared type: it is wrong even without running it, which is where my argument lies: if I can prove it to be wrong, I don't need to execute it at all.

asgrim: Yes, understood; it IS wrong, and tools say it's wrong, but a *consumer* of this API can still do the wrong thing. You are trusting that the docs are being read or the consumer is using static analysis, which is not always true. Right now we have excellent static analysis tools (psalm/phpstan) but they only analyse, never enforce. My issue is that by removing the assertions, people can do the wrong thing. IMO it goes against the core argument of your Extremely Defensive PHP talk - the whole poka yoke thing (avoid mistakes / fool proofing) - you can only plug it in one way. If you remove the assertions, there is more than one way to plug this thing in, and one/some/all of them are wrong. Just because it SAYS it's wrong, doesn't mean you can't actually do things wrong.

Until we get to a point where psalm/phpstan can actively enforce type declarations like `array<string, string>` or `callable () : Foo`, AND that we can be sure that all consumers use these tools to respect a defined API, I really strongly thing that runtime assertions should be kept.

composer.json
BackEndTea: What bizarre issues does it fail with? i'd like to fix those if possible

src/SourceLocator/StaticClassMapSourceLocator.php
Ocramius: > Yes, understood; it IS wrong, and tools say it's wrong, but a *consumer* of this API can still do the wrong thing.

Repeating something that we discussed in private, but I think this is a good foundation for dismissing issues as they come in. Much like "missing test case", "doesn't type-check" is also valid IMO.

 > IMO it goes against the core argument of your Extremely Defensive PHP talk - the whole poka yoke thing (avoid mistakes / fool proofing) - you can only plug it in one way.

Few things here:

 1. ["Extremely Defensive PHP"](https://github.com/Ocramius/extremely-defensive-php) is a talk from 2015, which I gave over 20 times at different conferences/events, and I still believe most of in, but it is very much in need for updates
 2. at the time that I wrote "Extremely Defensive PHP", tools like @phpstan and @vimeo /psalm and @doctrine /coding-standard were not existing, or in their infancy. @vimeo /psalm was first released in November 2016, @phpstan in July 2017, and I pushed @doctrine /coding-standard in January 2016
 3. the tools listed above supersede most of my (loudly touted) beliefs exposed in "Extremely Defensive PHP", and using them is very much equivalent to embracing [Poka-Yoke](https://en.wikipedia.org/wiki/Poka-yoke). It is always possible to skip them in some locations if you **really** know what you are doing, but things do indeed "click together" on "one true way of doing things"

Given the discussion so far, I think it is time for me to either design a new talk, or amend some stuff about it in a blogpost.

 > My issue is that by removing the assertions, people can do the wrong thing.

I agree that people are free to misuse the tool, but I am as free telling "this is not correct", and rejecting the issue (As mentioned above: "doesn't type-check" being equivalent to "cannot reproduce").

 > Until we get to a point where psalm/phpstan can actively enforce type declarations like `array<string, string>` or `callable () : Foo`, AND that we can be sure that all consumers use these tools to respect a defined API

This triggered a mental process on my end: as discussed in private, I could write a @vimeo /psalm plugin that acts on @composer `"post-install"` by running type-checks on any API surface that touches certain (configured) namespaces. I don't think that should affect this patch directly, but it is something that I will work on. It will be annoying as heck, but it will also be a massive improvement for the ecosystem, because opting out of type-checks will no longer be a viable choice.

 > I really strongly thing that runtime assertions should be kept.

I will see what I can do, but the current alternatives I see are:

 1. removing the added type declarations
 2. creating tests outside the `test/unit` (something like `test/api-misuse-errors`)

There is also the problem that assertions may fail to type-check for @phpstan or @vimeo /psalm as these tools catch up and start considering them (correctly) "redundant", which we will surely hit at some point, as type inference improves with the ecosystem tooling.

composer.json
Ocramius: @BackEndTea can you run https://github.com/Roave/BackwardCompatibilityCheck/pull/122/commits/0138f44bc17a235637e5c80511032061af349fdb `vendor/bin/infection`? It fails with a fatal error on a dynamic call from an extracted AST node.

The error message I got is in https://github.com/Roave/BackwardCompatibilityCheck/pull/122/commits/93788f8d5aaff5e8ceaf205f9d02a68bcc4e055b

BackEndTea: Ah, that looks like an error with anonymous/internal classes. Should be fixed in 0.13

src/SourceLocator/StaticClassMapSourceLocator.php
Ocramius: @asgrim now that I released `roave/you-are-using-it-wrong`, can we re-discuss this? Do we need to add the dependency, or are we fine with the patch as-is?

As mentioned above, I can either remove the type declarations overall, or add tests that are outside static analysis checking.

asgrim: Ship it I reckon - maybe make a separate patch to add `you-are-using-it-wrong`? :shipit:

composer.json
BackEndTea: @Ocramius With the release of 0.13, that version can be used now.

src/SourceLocator/StaticClassMapSourceLocator.php
Ocramius: Fair enuff

composer.json
Ocramius: Thanks @BackEndTea!

Ocramius:
Note: will need to chase down the latest mutation test results before merging.

Ocramius:
Expected failure:

```
[BC] REMOVED: Class Roave\BackwardCompatibility\SourceLocator\StubClassSourceLocator has been deleted
```

PR #142, opened 2019-07-17
#134 suppress exceptions in parsing `Fqsen` expressions for documented property types

### LINKED ISSUE #134: Do not crash when `Fqsen` parsing fails
As discussed with @sebastianbergmann (https://github.com/sebastianbergmann/phpunit/issues/3718), type annotations like `array<foo, bar>` and other more complex definitions currently lead to the phpdocumentor tooling to throw exceptions.

While it would be good to support these in full, this currently prevents picking up this tool at all in downstream consumers.

The suggestion is therefore to
 1. [x] Suppress `Fqsen` parsing errors
 2. [x] ~~Improve upstream `Fqsen` parsing tooling~~ not needed: see https://github.com/Roave/BackwardCompatibilityCheck/issues/143
 3. [x] ~~Adapt covariance/contravariance checks to support more complex types~~ moved to https://github.com/Roave/BackwardCompatibilityCheck/issues/143
 4. [x] ~~Extract covariance/contravariance checks to a new namespace/library that only does that, making it easier to upgrade its capabilities~~ moved to https://github.com/Roave/BackwardCompatibilityCheck/issues/143

Ocramius:
Fixes #134

### DIFF
--- a/src/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChanged.php
+++ b/src/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChanged.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Roave\BackwardCompatibility\DetectChanges\BCBreak\PropertyBased;
 
+use InvalidArgumentException;
 use Roave\BackwardCompatibility\Change;
 use Roave\BackwardCompatibility\Changes;
 use Roave\BackwardCompatibility\Formatter\ReflectionPropertyName;
@@ -34,8 +35,13 @@ final class PropertyDocumentedTypeChanged implements PropertyBased
             return Changes::empty();
         }
 
-        $fromTypes = array_unique($fromProperty->getDocBlockTypeStrings());
-        $toTypes   = array_unique($toProperty->getDocBlockTypeStrings());
+        try {
+            $fromTypes = array_unique($fromProperty->getDocBlockTypeStrings());
+            $toTypes   = array_unique($toProperty->getDocBlockTypeStrings());
+        } catch (InvalidArgumentException $failedToParseDocblock) {
+            // @TODO #134 improve docblock parsing upstream to remove this generic try-catch
+            return Changes::empty();
+        }
 
         sort($fromTypes);
         sort($toTypes);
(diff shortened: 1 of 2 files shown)


test/unit/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChangedTest.php
```diff
@@ -196,6 +206,7 @@ class TheClass {
             'privateDocblockToDifferentDocblock'                                => ['[BC] CHANGED: Type documentation for property TheClass#$privateDocblockToDifferentDocblock changed from int to float'],
             'duplicatePropertyTypesBeingDeduplicatedAreNotBcBreaks'             => [],
             'propertyTypeBeingDuplicatedAreNotBcBreaks'                         => [],
+            'propertyWithComplexDocblockThatCannotBeParsed'                     => [],
```
Ocramius: Currently, I'd say this is `mixed` -> `mixed`, so we shouldn't detect a BC break here.

src/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChanged.php
```diff
@@ -34,8 +35,13 @@ public function __invoke(ReflectionProperty $fromProperty, ReflectionProperty $t
             return Changes::empty();
         }
 
-        $fromTypes = array_unique($fromProperty->getDocBlockTypeStrings());
-        $toTypes   = array_unique($toProperty->getDocBlockTypeStrings());
+        try {
+            $fromTypes = array_unique($fromProperty->getDocBlockTypeStrings());
+            $toTypes   = array_unique($toProperty->getDocBlockTypeStrings());
+        } catch (InvalidArgumentException $failedToParseDocblock) {
+            // @TODO #134 improve docblock parsing upstream to remove this generic try-catch
```
asgrim: Not ideal to just hide it, but indeed we can't do much until we fix upstream, so this is kinda OK.

src/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChanged.php
```diff
@@ -34,8 +35,13 @@ public function __invoke(ReflectionProperty $fromProperty, ReflectionProperty $t
...
+            return Changes::empty();
```
asgrim: Do we have a way of returning "info" at all? or is it all just BC errors only?

Ocramius: I can return something, but then the build would be exiting with `1` here

asgrim: yeah, don't worry then :)

PR #158, opened 2019-09-18
#157 report tool version in cli

### LINKED ISSUE #157: --version does not show version
```
$ wget https://github.com/Roave/BackwardCompatibilityCheck/releases/download/4.0.0/roave-backward-compatibility-check.phar

$ php roave-backward-compatibility-check.phar --version
Console Tool
```

Ocramius:
Fixes #157

### DIFF
--- a/bin/roave-backward-compatibility-check.php
+++ b/bin/roave-backward-compatibility-check.php
@@ -7,6 +7,7 @@ namespace Roave\ApiCompareCli;
 use Composer\Factory;
 use Composer\Installer;
 use Composer\IO\ConsoleIO;
+use PackageVersions\Versions;
 use Roave\BackwardCompatibility\Command;
 use Roave\BackwardCompatibility\CompareClasses;
 use Roave\BackwardCompatibility\DetectChanges\BCBreak\ClassBased;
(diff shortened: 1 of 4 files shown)


bin/roave-backward-compatibility-check.php
```diff
@@ -51,7 +52,7 @@
         throw new RuntimeException('Could not find Composer autoload.php');
     })();
 
-    $application = new Application();
+    $application = new Application('roave/backward-compatibility-check', Versions::getVersion('roave/backward-compatibility-check'));
```
asgrim: does this also work in a `Phar` environment?

Ocramius: If the phar is built after `composer install` was performed, then yes. The `Versions` class will already have been generated at that point.

asgrim: I'd imagine it would be so :+1:

asgrim:
@Ocramius milestone `5.0.0`? could this not be `4.0.1`, as it's not really a BC break, or a feature (unless we already have BC breaks, or new features merged...?)

Ocramius:
@asgrim yeah, for now I'll keep it milestone `5.0.0`, but should we decide to split off `master` now, we can release `5.1.0` with everything that is in that milestone right now.

Ocramius:
Moved to 4.1.0, released.

PR #161, opened 2019-09-30
Feature/#159 remove version limitations from pahr assembly tool

### LINKED ISSUE #159: Remove version limitations from phar assembly tool
See https://github.com/Roave/BackwardCompatibilityCheck/blob/5092098994e3de84300a71c997494d6f637e839d/build-phar.sh#L7-L9

That is wrong, since:

 1. we will run the phar assembly on the correct version
 2. we shouldn't `composer update`, but rather `composer install --no-dev`
 3. script should completely bail out in case of errors

darenas31415:
Fixes #159

### DIFF
--- a/build-phar.sh
+++ b/build-phar.sh
@@ -2,16 +2,18 @@
 
 set -e
 
-mkdir -p tmp
+BOX_DIR="/tmp/box"
+
+mkdir -p ${BOX_DIR}
 
 # Install humbug/box
-composer --working-dir=tmp require humbug/box "^3.8" --no-interaction --no-progress --no-suggest
+composer --working-dir=${BOX_DIR} require humbug/box "^3.8" --no-interaction --no-progress --no-suggest
 
 # Remove dev dependencies for package distribution
 composer install --no-dev
 
-tmp/vendor/bin/box compile
+${BOX_DIR}/vendor/bin/box compile
 
 composer install
 
-rm -rf tmp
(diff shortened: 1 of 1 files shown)


build-phar.sh
```diff
@@ -1,14 +1,17 @@
...
 
-if [ ! -f box.phar ]; then
-    wget https://github.com/humbug/box/releases/download/3.0.0-beta.4/box.phar -O box.phar
-fi
+set -e
 
-# lock PHP to minimum allowed version
-composer config platform.php 7.2.0
-composer update --no-dev
+mkdir -p tmp
```
Ocramius: Probably better to not work in the CWD

darenas31415: What if we used a docker image with box to generate the phar? Something like https://github.com/jakzal/phpqa which comes with box installed

Ocramius: I'd generally prefer to have the latest version of box, not a tag from a derived docker image.

I think the patch is fine, but CWD needs to be `/tmp/something` rather than `./tmp` 👍

darenas31415: If that's OK with Travis CI (writing to /tmp) I can make the change. Some CIs don't let you go outside the working dir.

Ocramius: Yes, travis is pretty much allowing everything

Ocramius:
Thanks @darenas31415!

Issue #169, opened 2019-11-18
Phar for version 4.2.0


OndraM:
Hi, it seems phar file is missing from assets of release 4.2.0:
https://github.com/Roave/BackwardCompatibilityCheck/releases

Could you please add it?

Thanks!

asgrim:
@Ocramius hmm, this was the automated release, is there something missing there?

Ocramius:
Seems like the credentials aren't correct:
https://travis-ci.org/Roave/BackwardCompatibilityCheck/jobs/609657677

sebastianbergmann:
What is the status of this? Thanks!

Ocramius:
Inspecting the docs, I think this needs to be changed:

https://github.com/Roave/BackwardCompatibilityCheck/blob/026a96b4f523544b93f5d1228d83cd5303d7f976/.travis.yml#L29-L30

Specifically, the docs (https://docs.travis-ci.com/user/deployment/releases/) say that the format is:

```yml
deploy:
  provider: releases
  api_key: "GITHUB OAUTH TOKEN"
  file: "FILE TO UPLOAD"
  skip_cleanup: true
  on:
    tags: true
```

So we have a `secure:` too much there.

Ocramius:
Not solved: phar doesn't seem to be correctly built by travis, which reports `dist/roave-backward-compatibility-check.phar does not exist.`

Ocramius:
Now with phar: https://github.com/Roave/BackwardCompatibilityCheck/releases/tag/4.4.0

sebastianbergmann:
```
$ wget https://github.com/Roave/BackwardCompatibilityCheck/releases/download/4.4.0/roave-backward-compatibility-check.phar

$ php roave-backward-compatibility-check.phar --version
roave/backward-compatibility-check No version set (parsed as 1.0.0)@
```

Ocramius:
I think that's due to phar installation using `--no-scripts`: could be a bug in box?

PR #178, opened 2019-12-07
s/`deploy.api_key.secure`/`deploy.api_key`

Ocramius:
Ref: https://github.com/Roave/BackwardCompatibilityCheck/issues/169#issuecomment-562849108

Quoting:

> Inspecting the docs, I think this needs to be changed:
> 
> https://github.com/Roave/BackwardCompatibilityCheck/blob/026a96b4f523544b93f5d1228d83cd5303d7f976/.travis.yml#L29-L30
> 
> Specifically, the docs (https://docs.travis-ci.com/user/deployment/releases/) say that the format is:
> 
> ```yml
> deploy:
>   provider: releases
>   api_key: "GITHUB OAUTH TOKEN"
>   file: "FILE TO UPLOAD"
>   skip_cleanup: true
>   on:
>     tags: true
> ```

So we have a `secure:` too much there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,7 @@ git:
 
 deploy:
   provider: releases
-  api_key:
-    secure: $DEPLOY_TOKEN_SECURE
+  api_key: $DEPLOY_TOKEN_SECURE
   file: dist/roave-backward-compatibility-check.phar
   skip_cleanup: true
   on:


Ocramius:
@asgrim note: I don't know exactly how to test this, so we'll need to probably trigger a release to see if it works :\

PR #179, opened 2019-12-14
Upgrade nikolaposa/version to 4.0.0

nikolaposa:
### Proposed changes

Synchronize with changes made in Version 4.0.0 release.

- [x] fix `Version` and `VersionCollection` usage in accordance with their new API
- [x] fix assumption that input Version string prefix is not preserved when casting to string

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
         "ext-json": "*",
         "beberlei/assert": "^3.2.6",
         "composer/composer": "^1.9.0",
-        "nikolaposa/version": "@dev",
+        "nikolaposa/version": "^4.0.0",
         "ocramius/package-versions": "^1.5.1",
         "roave/better-reflection": "^3.5.0",
         "symfony/console": "^4.3.4",
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
         "ext-json": "*",
         "beberlei/assert": "^3.2.6",
         "composer/composer": "^1.9.0",
-        "nikolaposa/version": "^3.2.0",
+        "nikolaposa/version": "@dev",
         "ocramius/package-versions": "^1.5.1",
         "roave/better-reflection": "^3.5.0",
         "symfony/console": "^4.3.4",
--- a/src/Git/PickLastMinorVersionFromCollection.php
+++ b/src/Git/PickLastMinorVersionFromCollection.php
@@ -7,11 +7,11 @@ namespace Roave\BackwardCompatibility\Git;
 use Assert\Assert;
 use Symfony\Component\Process\Exception\LogicException;
 use Symfony\Component\Process\Exception\RuntimeException;
-use Version\Constraint\ComparisonConstraint;
-use Version\Constraint\CompositeConstraint;
-use Version\Constraint\ConstraintInterface;
+use Version\Comparison\Constraint\CompositeConstraint;
+use Version\Comparison\Constraint\Constraint;
+use Version\Comparison\Constraint\OperationConstraint;
 use Version\Version;
-use Version\VersionsCollection;
+use Version\VersionCollection;
 
 final class PickLastMinorVersionFromCollection implements PickVersionFromVersionCollection
 {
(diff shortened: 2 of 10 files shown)


nikolaposa:
When I already revealed that the new `nikolaposa/version` release of is near, are there any other things I could address in it?

Ocramius:
Patch seems good so far, except for a failure: https://travis-ci.org/Roave/BackwardCompatibilityCheck/jobs/624970650#L311

As for required additions in the upstream library, there's still this bug left: https://github.com/Roave/BackwardCompatibilityCheck/issues/37

In practice, we need to get both the version (normalized) and the original name of the symbol that led to that version. For example, `'release-1.2.3'` or `'v1.2.3'` or `'1.2.3'` are all equivalent, but we still want to know that a specific `Version` came from the string `'release-1.2.3'`.

This allows us to check out symbols by their original name.

nikolaposa:
Ok, I'll take a look at that failure.

As for `v1.2.3` issue, I always refer to: https://semver.org/#is-v123-a-semantic-version in conjunction with my library that tends to adhere to that semver document. So one way to solve this problem, while still keeping `Version` class within its original boundaries - object representing SemVer-compliant version number (v1.2.3 is not), is introducing a separate concept, such as:
```
class Tag
{
    private string $prefix;
    private Version $version;
}
```

I will consider some other solutions as well.

src/Git/PickLastMinorVersionFromCollection.php
```diff
@@ -38,13 +38,9 @@ public function assert(Version $version) : bool
...
         $matchingMinorVersions = $stableVersions
-            ->matching(new CompositeConstraint(
-                CompositeConstraint::OPERATOR_AND,
-                new ComparisonConstraint(ComparisonConstraint::OPERATOR_LTE, $lastVersion),
-                new ComparisonConstraint(
-                    ComparisonConstraint::OPERATOR_GTE,
-                    Version::fromString($lastVersion->getMajor() . '.' . $lastVersion->getMinor() . '.0')
-                )
+            ->matching(CompositeConstraint::and(
+                OperationConstraint::lessOrEqualTo($lastVersion),
```
nikolaposa: Static analysis check complains about possible null pointer error for `$lastVersion`. `VersionCollection->first()` indeed returns nullable type but there is an Assert at the beginning of this `forVersions()` method that makes sure that it deals only with non-empty collections.

I guess Psalm is not sophisticated enough to figure that out?

nikolaposa: I'll consider reverting `first()` and `last()` to raise "collection is empty" exception.

Ocramius: If the collection type is now allowing empty collections, it is correct that type analysis raises an issue here. An assertion on the result will be needed

nikolaposa: It always allowed. The only difference is that I decided to replace `CollectionIsEmpty` exception with nullable return type. Now I'm questioning that decision.

nikolaposa:
`Verify BC Breaks` stage is failing due to changes in `fromRepository()` method signature, caused by `VersionsCollection` -> `VersionCollection` rename.

nikolaposa:
For fixing https://github.com/Roave/BackwardCompatibilityCheck/issues/37, I can consider back-porting preserving of a Version string prefix in 3.2.x of my library.

Ocramius:
@nikolaposa no need - can happen in the next release

test/unit/Git/GetVersionCollectionFromGitRepositoryTest.php
```diff
@@ -76,6 +76,6 @@ public function testFromRepositoryAllowsVersionPrefix() : void
     {
         $this->makeTag('v1.0.0');
 
-        self::assertSame(['1.0.0'], $this->getTags());
+        self::assertSame(['v1.0.0'], $this->getTags());
```
Ocramius: This effectively fixes #37, right?

composer.json
```diff
@@ -6,7 +6,7 @@
         "ext-json": "*",
         "beberlei/assert": "^3.2.6",
         "composer/composer": "^1.9.0",
-        "nikolaposa/version": "^3.2.0",
+        "nikolaposa/version": "@dev",
```
Ocramius: To be changed before merge

test/unit/Git/GetVersionCollectionFromGitRepositoryTest.php
nikolaposa: I believe so, yes. That should be the scenario you referenced in other comment.

nikolaposa:
Sounds good!

Ocramius:
Moved to #199

PR #197, opened 2020-02-07
Fix Travis builds

### LINKED ISSUE #169: Phar for version 4.2.0

asgrim:
Fixes #169

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: php
 
-sudo: false
-
 php:
   - 7.3
   - 7.4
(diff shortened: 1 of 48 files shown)


asgrim:
Note: nightly build removed, if we definitely want to add it back in, we need to research how to allow failures on it *without* using the `matrix` directive (since it conflicts with `jobs` it seems)

Ocramius:
I'll probably be able to fix this up later tonight

Ocramius:
This was much more work than what I was hoping for, but it's green now @_@

asgrim:
> This was much more work than what I was hoping for, but it's green now @_@

That's why I left it for next week heh

PR #198, opened 2020-02-08
Do not skip cleanup when deploying

Ocramius:


.travis.yml
```diff
@@ -20,6 +20,8 @@ deploy:
   provider: releases
   api_key: $DEPLOY_TOKEN_SECURE
   file: dist/roave-backward-compatibility-check.phar
+  # do not skip cleanup: if the repo is cleaned up when deploying, then the phar is gone as well.
+  skip_cleanup: true
```
asgrim: Hmm, I read that this is deprecated and does nothing, the default is to skip cleanup now.

Ocramius: It was still affecting the build - in fact, 4.4.0 has a phar, and 4.3.0 doesn't :-\

asgrim: Humm I must've been mistaken then! Sorry, I removed it because Travis config checker was complaining about it heh. Maybe different build setup or something idk.

Ocramius: Not a problem: iz fixed nao :)

PR #201, opened 2020-02-12
Update github action example

jaapio:
Github actions checkout has been changed, it is now only fetching the current branch without any tags. Adding a few lines extra will help to make the action work again.


asgrim:
Thanks @jaapio !

Issue #31, opened 2018-04-15
Consider `final` in combination with `protected` to exclude some code analysis scenarios


Ocramius:
As discussed in person with @asgrim and @akrabat, the following change is not a BC break, because `final` already makes it impossible to rely on `protected` properties:

```php
final class Foo
{
    protected $bar = 'baz';
}
```

changed into:

```php
final class Foo
{
    private $bar = 'baz';
}
```

This kind of scenario is tricky to represent correctly with the current design of the `Comparator`

Ocramius:
Handled in #38

PR #202, opened 2020-02-14
Add test for property visibilty on final classes

jaapio:
This is a test case to show the issue we are facing with phpDocumentor. Both cases are failing now, but I would expect them to pass.

### DIFF
--- a/test/unit/DetectChanges/BCBreak/ClassBased/PropertyChangedTest.php
+++ b/test/unit/DetectChanges/BCBreak/ClassBased/PropertyChangedTest.php
@@ -80,4 +80,91 @@ PHP
             )
         );
     }
+
+    /**
+     * @dataProvider finalClassProvider
+     */
+    public function testWillIgnoreChangeInFinalClass($fromLocator, $toLocator) : void
+    {
+
+
+        $comparator = $this->createMock(PropertyBased::class);
+
+        $comparator
+            ->expects(self::exactly(2))
+            ->method('__invoke')
+            ->willReturnCallback(static function (ReflectionProperty $from, ReflectionProperty $to) : Changes {
+                $propertyName = $from->getName();
+
+                self::assertSame($propertyName, $to->getName());
+
+                return Changes::fromList(Change::added($propertyName, true));
+            });
+
+        Assertion::assertChangesEqual(
+            Changes::empty(),
+            (new PropertyChanged($comparator))->__invoke(
+                (new ClassReflector($fromLocator))->reflect('TheClass'),
+                (new ClassReflector($toLocator))->reflect('TheClass')
+            )
+        );
+    }
+
+    public function finalClassProvider() : array
+    {
+        $astLocator = (new BetterReflection())->astLocator();
+
+        return [
+            'final class' => [
+                new StringSourceLocator(
+                    <<<'PHP'
+<?php
+
+final class TheClass {
+    protected $b;
+}
+PHP
+                    ,
+                    $astLocator
+                ),
+                new StringSourceLocator(
+                    <<<'PHP'
+<?php
+final class TheClass {
+    private $b;
+}
+PHP
+                    ,
+                    $astLocator
+                ),
+            ],
+            'final class extending baseclass' => [
+                new StringSourceLocator(
+                    <<<'PHP'
+<?php
+abstract class baseClass {}
+
+final class TheClass extends baseClass {
+    protected $b;
+}
+PHP
+                    ,
+                    $astLocator
+                ),
+                new StringSourceLocator(
+                    <<<'PHP'
+<?php
+abstract class baseClass {}
+
+final class TheClass extends baseClass{
+    private $b;
+}
+PHP
+                    ,
+                    $astLocator
+                ),
+            ],
+        ];
+
+    }
 }


test/unit/DetectChanges/BCBreak/ClassBased/PropertyChangedTest.php
```diff
@@ -80,4 +80,91 @@ class TheClass {
...
+}
+PHP
+                    ,
+                    $astLocator
+                ),
+                new StringSourceLocator(
+                    <<<'PHP'
+<?php
+final class TheClass {
+    private $b;
```
Ocramius: In this case, I'd say that the property was indeed changed: this shouldn't produce a `Changes::empty()`

test/unit/DetectChanges/BCBreak/ClassBased/PropertyChangedTest.php
```diff
@@ -80,4 +80,91 @@ class TheClass {
...
+                    ,
+                    $astLocator
+                ),
+                new StringSourceLocator(
+                    <<<'PHP'
+<?php
+abstract class baseClass {}
+
+final class TheClass extends baseClass{
+    private $b;
```
Ocramius: Similar here, property `$b` was indeed changed, but not on `baseClass`

Ocramius:
@jaapio what's the original error message that started this investigation? Maybe we can go in more detail in there, or maybe it's more of a mis-configuration issue (for `final` classes)

test/unit/DetectChanges/BCBreak/ClassBased/PropertyChangedTest.php
```diff
@@ -80,4 +80,91 @@ class TheClass {
...
+
+        $comparator
+            ->expects(self::exactly(2))
+            ->method('__invoke')
+            ->willReturnCallback(static function (ReflectionProperty $from, ReflectionProperty $to) : Changes {
+                $propertyName = $from->getName();
+
+                self::assertSame($propertyName, $to->getName());
+
+                return Changes::fromList(Change::added($propertyName, true));
```
Ocramius: The property wasn't added, since it is present in both `$from` and `$to`

jaapio:
The build which was failing can be found here: 

https://github.com/phpDocumentor/ReflectionDocBlock/runs/442034844?check_suite_focus=true

The invalid error is in my opinion here:

```
[BC] REMOVED: Property phpDocumentor\Reflection\DocBlock\Tags\PropertyRead#$variableName was removed
```
On line: https://github.com/phpDocumentor/ReflectionDocBlock/pull/207/files#diff-788f8ed7c572c70343f0e841def48b5cL37

Protected properties in a final class do not make sense. Like implemented in this issue: https://github.com/Roave/BackwardCompatibilityCheck/issues/31

test/unit/DetectChanges/BCBreak/ClassBased/PropertyChangedTest.php
jaapio: Why? it doesn't make sense to have a protected property in this case. nobody can access it only the current class, so this is not a bc break. since private and protected are equal in this case. Since there is no parent class.

test/unit/DetectChanges/BCBreak/ClassBased/PropertyChangedTest.php
jaapio: Because `$b` is not used in `baseClass` it could not be detected as a change.

test/unit/DetectChanges/BCBreak/ClassBased/PropertyChangedTest.php
Ocramius: `PropertyChanged` forwards checks to futher BCBreak checkers: it doesn't really raise a BC break itself.

In the config of this project, changes to `final` classes are scoped within the `FinalClassChanged` checks:

https://github.com/Roave/BackwardCompatibilityCheck/blob/3dd77835db29c46e571c41acedf733766015b871/bin/roave-backward-compatibility-check.php#L182

I think what's being hit is this other checker: 

https://github.com/Roave/BackwardCompatibilityCheck/blob/3dd77835db29c46e571c41acedf733766015b871/bin/roave-backward-compatibility-check.php#L83-L90

As you can see, not scoped within `final` classes context: we probably want such a check to only apply to open classes, or to `public` properties of `final` classes, so the rule has to be nested a bit deeper.

test/unit/DetectChanges/BCBreak/ClassBased/PropertyChangedTest.php
Ocramius: Similar to discussion at https://github.com/Roave/BackwardCompatibilityCheck/pull/202#discussion_r379490933: I think the bug is not in the sources themselves, but rather in the configuration. 

Also see `Roave\BackwardCompatibility\DetectChanges\BCBreak\ClassBased\PropertyRemoved`: that's the one raising this specific warning.

Ocramius:
> It could be that the test cases are not completely valid, but they do illustrate the issue. Please let me know if it required another test. I can imagine that you are filtering the results when the error class is final?

I think our best bet for now is to change the tests you wrote to target `Roave\BackwardCompatibility\DetectChanges\BCBreak\ClassBased\PropertyRemoved`, which is more specific to the error you are getting right now.

Ocramius:
@jaapio I re-visited this while looking at the backlog, and isolated it to #253. 

Sorry it took so long: 'been doing very little OSS due to work commitments.

jaapio:
Thanks for addressing this!

PR #247, opened 2020-06-19
Fix build

Jean85:
This should make the build green.

Travis config edits are meant to silence config warnings: https://travis-ci.org/github/Roave/BackwardCompatibilityCheck/builds/696432282/config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,10 @@ git:
 
 deploy:
   provider: releases
-  api_key: $DEPLOY_TOKEN_SECURE
+  token: $DEPLOY_TOKEN_SECURE
   file: dist/roave-backward-compatibility-check.phar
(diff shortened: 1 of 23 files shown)


.travis.yml
```diff
@@ -18,10 +18,10 @@ git:
 
 deploy:
   provider: releases
-  api_key: $DEPLOY_TOKEN_SECURE
+  token: $DEPLOY_TOKEN_SECURE
   file: dist/roave-backward-compatibility-check.phar
   # do not skip cleanup: if the repo is cleaned up when deploying, then the phar is gone as well.
-  skip_cleanup: true
+  cleanup: false
```
Ocramius, line 24: This stuff really needs XSDs 😭

Jean85:
Fixed now, I hope. CS automatic fixes were clashing with SCA so I had to do a few iterations; good news is that SCA got better and I just had to get rid of a couple of type annotations.

PR #254, opened 2020-06-22
Moved all builds and tests to github actions

Ocramius:
Moved all builds and tests to github actions

Note: this requires `composer/composer:^2@dev`, which will be a problem for downstream consumers that have the default minimum stability: will make a new release once `composer/composer:^2` goes GA.


.github/workflows/backward-compatibility-check.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: "Cache dependencies"
+        uses: "actions/cache@v2"
+        with:
+          path: |
+            ~/.composer/cache
+            vendor
+          key: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}"
+          restore-keys: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}"
+
+      - name: "Install lowest dependencies"
```
Ocramius, line 48: Note: while most of this is skipped, I kept the duplication across the file to easier diff them between each other.

This is hopefully (at some point) solved by github having a better DSL for github actions: YAML sucks badly for all this stuff, but there's no way around hipster culture, as it seems.

Ocramius:
Actions don't yet support PHP 7.4.7 - only 7.4.6 currently available.

This will need to happen on another day :-)

PR #261, opened 2020-07-17
Replace Travis-CI with Github Actions

Ocramius:
Note: this upgrades the package to use `composer/composer:^2.0@dev`: that will make this version not installable for a while, until there's a stable `composer/composer:^2` release


Ocramius:
Merged into #254

.github/workflows/backwards-compatibility.yml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+        with:
+          fetch-depth: 0
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "pcov"
```
DaveLiddament, line 31: I don't think there is any need for coverage tool here: See https://github.com/marketplace/actions/setup-php-action#disable-coverage

Similar comment applies to all the other workflows apart from the test one.

.github/workflows/backwards-compatibility.yml
```diff
@@ -0,0 +1,58 @@
...
+          php-version: "${{ matrix.php-version }}"
+          ini-values: memory_limit=-1
+          tools: composer:v2, cs2pr
+
+      - name: "Cache dependencies"
+        uses: "actions/cache@v2"
+        with:
+          path: |
+            ~/.composer/cache
+            vendor
```
DaveLiddament, line 41: Also someone reviewed my github actions, which I copied from here. They said this. I've not got to the bottom of it yet, but it might be of interest to you though...

You shouldn't cache vendor/, see https://github.com/shivammathur/setup-php#cache-composer-dependencies

Please do not cache vendor directory using action/cache as that will have side-effects.

.github/workflows/backwards-compatibility.yml
Ocramius: I kept this to keep the files as similar as possible, allowing for easy diffing between each other.

Until github actions allow imports of external configurations, it makes sense to keep them close to ease maintenance.

.github/workflows/backwards-compatibility.yml
Ocramius: Considering that the cache depends on `composer.lock`, the only side-effect in here may come from `ocramius/package-versions`, which has very minimal issues on this front?

PR #266, opened 2020-08-17
Detect BC breaks due to parameter name changes, in line with PHP 8.0 named arguments requirements

### LINKED ISSUE #264: PHP8: Named parameters support & opt-out attribute
Since it looks like [named parameters](https://wiki.php.net/rfc/named_params) will pass, with no proposal of having a way to opt out of this feature, I'd like to propose introducing an attribute to have parameter names excluded from BCC rules, by way of annotation.

The main comment I've seen in support of named parameters is that "you can just say you're not supporting named parameters". Unfortunately, just saying that (e.g. by writing in your README.md) doesn't make tools understand that fact, so we need to recognise this some other way. The RFC even goes as far as to mention an "opt-out" possibility, but it never expands on it beyond suggesting an annotation such as `<<NoNamedArgs>>`. Therefore, unless another RFC is proposed to add this, we can be fairly sure that there will be no runtime checks to prevent calling a method with named-parameter syntax.

So: onto the proposal: Add an annotation to method definition that says "you must not call this method using named parameter syntax". Static analysers such as Psalm/PHPStan could check this at call-time: I believe Psalm "may" support this if there is demand, and PHPStan are actively looking to support this kind of annotation.

Because there is chance for BC break by renaming parameters with the named param RFC (even if a rename is rare), we will need to add this as a rule now from any PHP 8 code. However, we can use the above annotation to exclude any parameter renames from being marked as BC breaks within this tool.

Before API:
```
function supportsNamedParams($a, $b, $c);

@@NoNamedParams
function doesNotSupportNamedParams($a, $b, $c);
```

After API
```
function supportsNamedParams($d, $e, $f); // would be marked as a BC break due to named params in PHP 8

@@NoNamedParams
function doesNotSupportNamedParams($d, $e, $f); // param renames are ignored due to annotation, not classed as a BC break
```

asgrim:
Fixes #264

### DIFF
--- a/src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
+++ b/src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BackwardCompatibility\DetectChanges\BCBreak\FunctionBased;
+
+use Roave\BackwardCompatibility\Change;
+use Roave\BackwardCompatibility\Changes;
+use Roave\BackwardCompatibility\Formatter\ReflectionFunctionAbstractName;
+use Roave\BetterReflection\Reflection\ReflectionFunctionAbstract;
+use Roave\BetterReflection\Reflection\ReflectionParameter;
+
+use function array_intersect_key;
+use function Safe\sprintf;
+
+/**
+ * @todo only apply this to PHP 8+ code
+ *
+ * Detects a change in a parameter type
+ *
+ * This is mostly useful for methods, where a change in a parameter name is not allowed in
+ * inheritance/interface scenarios, except if annotated with `no-named-arguments`
+ */
+final class ParameterNameChanged implements FunctionBased
+{
+    private ReflectionFunctionAbstractName $formatFunction;
+
+    public function __construct()
+    {
+        $this->formatFunction = new ReflectionFunctionAbstractName();
+    }
+
+    public function __invoke(ReflectionFunctionAbstract $fromFunction, ReflectionFunctionAbstract $toFunction): Changes
+    {
+        return Changes::fromIterator($this->checkSymbols(
+            $fromFunction->getParameters(),
+            $toFunction->getParameters()
+        ));
+    }
+
+    /**
+     * @param ReflectionParameter[] $from
+     * @param ReflectionParameter[] $to
+     *
+     * @return iterable|Change[]
+     */
+    private function checkSymbols(array $from, array $to): iterable
+    {
+        foreach (array_intersect_key($from, $to) as $index => $commonParameter) {
+            yield from $this->compareParameter($commonParameter, $to[$index]);
+        }
+    }
+
+    /**
+     * @return iterable|Change[]
+     */
+    private function compareParameter(ReflectionParameter $fromParameter, ReflectionParameter $toParameter): iterable
+    {
+        // @todo detect if the method has a @no-named-arguments annotation, and return if so
+
+        $fromName = $fromParameter->getName();
+        $toName   = $toParameter->getName();
+
+        if ($fromName === $toName) {
+            return;
+        }
+
+        yield Change::changed(
+            sprintf(
+                'Parameter %d of %s changed name from %s to %s',
+                $fromParameter->getPosition(),
+                $this->formatFunction->__invoke($fromParameter->getDeclaringFunction()),
+                $fromName,
+                $toName
+            ),
+            true
+        );
+    }
+}
--- a/src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
+++ b/src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
@@ -16,7 +16,8 @@ use function Safe\sprintf;
 /**
  * @todo only apply this to PHP 8+ code
  *
- * Detects a change in a parameter type
+ * Detects a change in a parameter name, which must now be considered a BC break as of PHP 8 (specifically, since the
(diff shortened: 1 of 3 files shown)


src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
```diff
@@ -0,0 +1,79 @@
...
+use Roave\BetterReflection\Reflection\ReflectionFunctionAbstract;
+use Roave\BetterReflection\Reflection\ReflectionParameter;
+
+use function array_intersect_key;
+use function Safe\sprintf;
+
+/**
+ * @todo only apply this to PHP 8+ code
+ *
+ * Detects a change in a parameter type
```
asgrim, line 19: Fix docblocks from C&P

Ocramius:
Note: also needs to configure the new rule in `bin/*.php`

src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
```diff
@@ -0,0 +1,79 @@
...
+            yield from $this->compareParameter($commonParameter, $to[$index]);
+        }
+    }
+
+    /**
+     * @return iterable|Change[]
+     */
+    private function compareParameter(ReflectionParameter $fromParameter, ReflectionParameter $toParameter): iterable
+    {
+        // @todo detect if the method has a @no-named-arguments annotation, and return if so
```
asgrim, line 59: `@todo`

src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
```diff
@@ -0,0 +1,79 @@
...
+use Roave\BackwardCompatibility\Changes;
+use Roave\BackwardCompatibility\Formatter\ReflectionFunctionAbstractName;
+use Roave\BetterReflection\Reflection\ReflectionFunctionAbstract;
+use Roave\BetterReflection\Reflection\ReflectionParameter;
+
+use function array_intersect_key;
+use function Safe\sprintf;
+
+/**
+ * @todo only apply this to PHP 8+ code
```
asgrim, line 17: `@todo`

src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
```diff
@@ -0,0 +1,119 @@
...
+                $fromParameter->getPosition(),
+                $this->formatFunction->__invoke($fromParameter->getDeclaringFunction()),
+                $fromName,
+                $toName
+            ),
+            true
+        );
+    }
+
+    private function methodHasNoNamedArgumentsAnnotation(ReflectionFunctionAbstract $function): bool
```
asgrim, line 115: I wonder if any static analysis library will add some kind of PHP attribute like `#[NoNamedAttributes]` or something, we could look for that too, although I think we'd need to add support upstream in Better Reflection for attributes first, so that could be an improvement to come later.

src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
asgrim: I think the only way we can realistically do this is to:

 - Write a class like `ReadPhpVersionRequirementsFromComposerJson` that:
   - reads `composer.json`
   - find the `require.php` key and stores the min/max value
   - it would be run lazily
 - The class can be constructor-injected into the `ParameterNamedChanged` check
 - The `ParameterNamedChanged` can ask the class if we intersect PHP 8+ at all, if not, return

Ocramius: IMO this also applies to PHP 7.x code - running the same code on 8.x will have these issues, regardless if we're developing on 7.x right now

src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
Ocramius: Possibly to be done once the ecosystem agrees on these annotations - too early for us to jump in on this

src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * @param ReflectionParameter[] $from
+     * @param ReflectionParameter[] $to
+     *
+     * @return iterable|Change[]
+     */
+    private function checkSymbols(array $from, array $to): iterable
+    {
+        foreach (array_intersect_key($from, $to) as $index => $commonParameter) {
```
Ocramius, line 86: One test is missing here: specifically something like:

```diff
-function foo($a, $b, $c){}
+function foo($b, $c){}
```

This particular BCBreak detector should not pick that up

Ocramius (changes requested):

src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
Ocramius: 4fa0de2315627bfaccce94a1aa05dc7b31ef4e34

src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php
Ocramius: 17adbea4322a7eddbc19f59e14828c5f69eac9a6

Ocramius (approved):
@asgrim I've adapted your patch to latest `6.0.x` state, and it's good to go :ship:

asgrim:
Thanks @Ocramius !

PR #269, opened 2020-11-07
Improve message for classes with removed traits with properties

acrobat:
[This PR](https://github.com/KnpLabs/php-github-api/pull/924/files#diff-8d9f29fab8b1d22b8f1b92a0be543d4b8e68ac2d475f7c3ed8da184ade8ad91f) removed a trait use from a class and the checker correctly indicated a BC break but the message was a bit confusing. So I've improved the output message for this case and added a few tests to cover this use case.

Received message:
```
[BC] REMOVED: Property Github\Api\AcceptHeaderTrait#$acceptHeaderValue was removed
```

Expected mesage:
```
[BC] REMOVED: Property Github\Api\Repository\Checks#$acceptHeaderValue was removed
```

### DIFF
--- a/src/Formatter/ReflectionPropertyName.php
+++ b/src/Formatter/ReflectionPropertyName.php
@@ -10,10 +10,16 @@ final class ReflectionPropertyName
 {
     public function __invoke(ReflectionProperty $property): string
     {
-        if ($property->isStatic()) {
-            return $property->getDeclaringClass()->getName() . '::$' . $property->getName();
+        $declaringClass = $property->getDeclaringClass();
+        $className      = $declaringClass->getName();
+        if ($declaringClass->isTrait()) {
+            $className = $property->getImplementingClass()->getName();
         }
 
-        return $property->getDeclaringClass()->getName() . '#$' . $property->getName();
+        if ($property->isStatic()) {
+            return $className . '::$' . $property->getName();
+        }
+
+        return $className . '#$' . $property->getName();
     }
 }
--- a/test/unit/DetectChanges/BCBreak/ClassBased/PropertyRemovedTest.php
+++ b/test/unit/DetectChanges/BCBreak/ClassBased/PropertyRemovedTest.php
@@ -100,6 +100,71 @@ PHP
                     '[BC] REMOVED: Property FinalClass#$removedPublicProperty was removed',
                 ],
             ],
+            'removed trait use from class' => [
+                (new ClassReflector(new StringSourceLocator(
+                    <<<'PHP'
+<?php
+
+trait PropertyTrait {
+    protected $testProperty;
+}
+
+class TestClass
+{
+    use PropertyTrait;
+}
+PHP
+                    ,
+                    $locator
+                )))->reflect('TestClass'),
+                (new ClassReflector(new StringSourceLocator(
+                    <<<'PHP'
+<?php
+
+class TestClass
+{
+}
+PHP
+                    ,
+                    $locator
+                )))->reflect('TestClass'),
+                ['[BC] REMOVED: Property TestClass#$testProperty was removed'],
+            ],
+            'removed property from trait' => [
+                (new ClassReflector(new StringSourceLocator(
+                    <<<'PHP'
+<?php
+
+trait PropertyTrait {
+    protected $testProperty;
+}
+
+class TestClass
+{
+    use PropertyTrait;
+}
+PHP
+                    ,
+                    $locator
+                )))->reflect('TestClass'),
+                (new ClassReflector(new StringSourceLocator(
+                    <<<'PHP'
+<?php
+
+trait PropertyTrait {
+    
+}
+
+class TestClass
+{
+    use PropertyTrait;
+}
+PHP
+                    ,
+                    $locator
+                )))->reflect('TestClass'),
+                ['[BC] REMOVED: Property TestClass#$testProperty was removed'],
+            ],
         ];
     }
 }
--- a/test/unit/Formatter/ReflectionPropertyNameTest.php
+++ b/test/unit/Formatter/ReflectionPropertyNameTest.php
@@ -41,9 +41,29 @@ final class ReflectionPropertyNameTest extends TestCase
 <?php
 
 namespace {
+    trait TestTrait {
+        public static $b;
+        public $c;
+    }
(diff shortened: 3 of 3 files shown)


src/Formatter/ReflectionPropertyName.php
```diff
@@ -10,10 +10,15 @@ final class ReflectionPropertyName
 {
     public function __invoke(ReflectionProperty $property): string
     {
+        $className = $property->getDeclaringClass()->getName();
```
Ocramius, line 13: Avoid repeated declaring class calls, if possible

test/unit/DetectChanges/BCBreak/ClassBased/TraitUsePropertyRemovedTest.php
```diff
@@ -0,0 +1,123 @@
+<?php
```
Ocramius, line 1: There should be a pre-existing unit test for this: can you check? In theory, we just need to improve that

Ocramius (changes requested):

test/unit/DetectChanges/BCBreak/ClassBased/TraitUsePropertyRemovedTest.php
acrobat: I didn't find a class based test with a trait use, also changing the message reported didn't cause any failing tests so I expected that there was no test in the passed which covered this behaviour

src/Formatter/ReflectionPropertyName.php
acrobat: I've replaced the repeated `getDeclaringClass` calls by a single call

test/unit/DetectChanges/BCBreak/ClassBased/TraitUsePropertyRemovedTest.php
Ocramius: The class is https://github.com/Roave/BackwardCompatibilityCheck/blob/2eae3f72084f0bdc59847e486581f7882ba6a88c/test/unit/Formatter/ReflectionPropertyNameTest.php#L22 - the tests generally use a `StringSourceLocator` to make code snippets small and isolated :)

test/unit/DetectChanges/BCBreak/ClassBased/TraitUsePropertyRemovedTest.php
```diff
@@ -0,0 +1,118 @@
...
+use Roave\BackwardCompatibility\DetectChanges\BCBreak\ClassBased\PropertyRemoved;
+use Roave\BetterReflection\BetterReflection;
+use Roave\BetterReflection\Reflection\ReflectionClass;
+use Roave\BetterReflection\Reflector\ClassReflector;
+use Roave\BetterReflection\SourceLocator\Type\StringSourceLocator;
+
+use function array_map;
+use function iterator_to_array;
+
+final class TraitUsePropertyRemovedTest extends TestCase
```
Ocramius, line 18: I think this should be part of the `PropertyRemovedTest` (discussed in https://github.com/Roave/BackwardCompatibilityCheck/pull/269/files#r519199158)

test/unit/DetectChanges/BCBreak/TraitBased/PropertyRemovedTest.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+declare(strict_types=1);
+
+namespace RoaveTest\BackwardCompatibility\DetectChanges\BCBreak\TraitBased;
```
Ocramius, line 5: There is no `TraitBased\PropertyRemoved` - probably why this was so confusing

Ocramius (changes requested):
Mutation tests report this mutation:

```

-        if ($declaringClass->isTrait() && $property->getImplementingClass()->getName() !== $declaringClass->getName()) {
+        if ($declaringClass->isTrait() || $property->getImplementingClass()->getName() !== $declaringClass->getName()) {
             $className = $property->getImplementingClass()->getName();
         }
```

We're probably missing an inheritance-related test to cover that

test/unit/DetectChanges/BCBreak/ClassBased/TraitUsePropertyRemovedTest.php
acrobat: I've added some testcases to that test

test/unit/DetectChanges/BCBreak/ClassBased/TraitUsePropertyRemovedTest.php
acrobat: I've removed this file and moved the tests to the propertyRemoved test file

acrobat:
@Ocramius I don't know how to fix the mutation tests error, any pointers on how to fix this? As I would've expected that the other added tests would cover this behaviour change 🤔

Ocramius:
The mutation is:

```
 -        if ($declaringClass->isTrait() && $property->getImplementingClass()->getName() !== $declaringClass->getName()) {
+        if ($declaringClass->isTrait() || $property->getImplementingClass()->getName() !== $declaringClass->getName()) {
             $className = $property->getImplementingClass()->getName();
         }
```

Perhaps the second part of the conditional is never needed at all? Alternatively, it could be only valid in complex inheritance scenarios, such as when a property exists, but is replaced by an imported trait.

acrobat:
@Ocramius You are actually right, we just need to use the implementing class when formatting a trait. If it is a error for the trait itself the implementing class will reference the trait, if the trait is used in a class it will reference the class.

I've added some extra case to verify this and every is ok now 👌

test/unit/DetectChanges/BCBreak/TraitBased/PropertyRemovedTest.php
```diff
@@ -0,0 +1,84 @@
+<?php
```
Ocramius, line 1: Can we drop this before merging? In theory, all changes are covered by the `ReflectionPropertyNameTest`

acrobat: Ok, I've removed the test!

Ocramius (approved):
Excellent, thank you, @acrobat!

PR #306, opened 2021-05-19
replace `thecodingmachine/safe`, `webmozart/assert`, and `symfony/process` by `azjezz/psl`

azjezz:
- replace `thecodingmachine/safe` by `azjezz/psl`
- replace `webmozart/assert` by `azjezz/psl`
- replace `symfony/process` by `azjezz/psl` 
- upgraded psalm from v3 to v4
- upgraded phpunit to 9.5 ( + migrated configurations accordingly )
- replaced usage of PHP builtin array functions by `Psl\Vec`, and `Psl\Dict` components.


azjezz:
@Ocramius workflow passes in my branch ( https://github.com/azjezz/BackwardCompatibilityCheck/actions ), but needs approval here to run.

psalm.xml
```diff
@@ -19,5 +17,22 @@
...
+                <directory name="test"/>
+                <directory name="bin"/>
+            </errorLevel>
+        </DeprecatedClass>
+
+        <DeprecatedMethod>
+            <errorLevel type="suppress">
+                <directory name="test/"/>
+            </errorLevel>
+        </DeprecatedMethod>
```
azjezz, line 32: PackageVersions and Test::at($n) are the sources of deprecation

composer.json
```diff
@@ -27,10 +25,11 @@
     "require-dev": {
         "doctrine/coding-standard": "^8.2.0",
         "infection/infection": "^0.16.4",
-        "phpunit/phpunit": "^9.2.6",
+        "php-standard-library/psalm-plugin": "^1.1",
+        "phpunit/phpunit": "^9.5",
         "roave/security-advisories": "dev-master",
         "squizlabs/php_codesniffer": "^3.6.0",
-        "vimeo/psalm": "^3.12.2"
+        "vimeo/psalm": "^4.7"
```
Ocramius, line 32: Should probably be `^4.7.2`

bin/roave-backward-compatibility-check.php
```diff
@@ -43,7 +43,9 @@
 
         foreach ($autoloaderLocations as $autoload) {
             if (file_exists($autoload)) {
-                /** @noinspection PhpIncludeInspection */
+                /**
```
Ocramius, line 46: Can probably be dropped

src/CompareClasses.php
```diff
@@ -40,16 +38,16 @@ public function __invoke(
         ClassReflector $pastSourcesWithDependencies,
         ClassReflector $newSourcesWithDependencies
     ): Changes {
-        $definedApiClassNames = array_map(
-            static function (ReflectionClass $class): string {
-                return $class->getName();
-            },
-            array_filter(
+        $definedApiClassNames = Dict\map(
```
Ocramius, line 41: Is there an advantage in using `Dict\map` over `array_map()` here? Here, the structure is not an iterable, so it would be better to rely on engine core concepts

Ocramius: Same goes for `array_filter()`

src/DetectChanges/BCBreak/ClassBased/AncestorRemoved.php
```diff
@@ -21,20 +20,20 @@ final class AncestorRemoved implements ClassBased
 {
     public function __invoke(ReflectionClass $fromClass, ReflectionClass $toClass): Changes
     {
-        $removedAncestors = array_merge(
```
Ocramius, line 24: Similar to above: would suggest keeping `array_merge()` and `array_diff()` here, since they are well known engine constructs that are as-safe-as PSL here?

src/DetectChanges/BCBreak/ClassBased/ConstantChanged.php
```diff
@@ -38,7 +37,7 @@ public function __invoke(ReflectionClass $fromClass, ReflectionClass $toClass):
      */
     private function checkSymbols(array $from, array $to): iterable
     {
-        foreach (array_keys(array_intersect_key($from, $to)) as $name) {
+        foreach (Vec\keys(Dict\intersect_by_key($from, $to)) as $name) {
```
Ocramius, line 40: Similar to other comments about `array_keys()` and `array_intersect_key()`, I would suggest keeping this as it was

src/DetectChanges/BCBreak/ClassBased/ConstantRemoved.php
```diff
@@ -4,38 +4,35 @@
...
-use function array_map;
-use function array_values;
-use function Safe\sprintf;
-
 final class ConstantRemoved implements ClassBased
 {
     public function __invoke(ReflectionClass $fromClass, ReflectionClass $toClass): Changes
     {
-        $removedConstants = array_diff_key(
+        $removedConstants = Dict\diff_by_key(
```
Ocramius, line 19: Similar to other comments about `array_keys()` and `array_intersect_key()`, I would suggest keeping this as it was

src/DetectChanges/BCBreak/ClassBased/MethodChanged.php
```diff
@@ -38,7 +35,7 @@ public function __invoke(ReflectionClass $fromClass, ReflectionClass $toClass):
      */
     private function checkSymbols(array $from, array $to): iterable
     {
-        foreach (array_keys(array_intersect_key($from, $to)) as $name) {
+        foreach (Vec\keys(Dict\intersect_by_key($from, $to)) as $name) {
```
Ocramius, line 38: Similar to other comments about `array_keys()` and `array_intersect_key()`, I would suggest keeping this as it was

src/DetectChanges/BCBreak/ClassBased/MethodChanged.php
```diff
@@ -48,10 +45,10 @@ private function methods(ReflectionClass $class): array
     {
         $methods = $class->getMethods();
 
-        return array_combine(
-            array_map(static function (ReflectionMethod $method): string {
-                return strtolower($method->getName());
-            }, $methods),
+        return Dict\associate(
+            Vec\map($methods, static function (ReflectionMethod $method): string {
+                return Str\lowercase($method->getName());
```
Ocramius, line 50: `strtolower()` worked OK here - why replace it? :thinking:

src/DetectChanges/BCBreak/ClassBased/MethodChanged.php
```diff
@@ -48,10 +45,10 @@ private function methods(ReflectionClass $class): array
     {
         $methods = $class->getMethods();
 
-        return array_combine(
-            array_map(static function (ReflectionMethod $method): string {
-                return strtolower($method->getName());
-            }, $methods),
+        return Dict\associate(
+            Vec\map($methods, static function (ReflectionMethod $method): string {
```
Ocramius, line 49: Similar to other comments about `array_map()`, I would suggest keeping this as it was

src/LocateDependencies/LocateDependenciesViaComposer.php
```diff
@@ -38,12 +37,10 @@ public function __construct(
...
     public function __invoke(string $installationPath): SourceLocator
     {
-        Assert::file($installationPath . '/composer.json');
+        Psl\invariant(Filesystem\is_file($installationPath . '/composer.json'), 'Could not locate composer.json within installation path.');
 
         $this->runInDirectory(function () use ($installationPath): void {
-            $installer = ($this->makeComposerInstaller)($installationPath);
-
-            assert($installer instanceof Installer);
+            $installer = Type\object(Installer::class)->assert(($this->makeComposerInstaller)($installationPath));
```
Ocramius, line 43: I think `$this->makeComposerInstaller` is missing an `@psalm-var callable(): Installer`? This assertion shouldn't be needed then...

src/Support/ArrayHelpers.php
```diff
@@ -1,28 +0,0 @@
...
-declare(strict_types=1);
-
-namespace Roave\BackwardCompatibility\Support;
-
-use InvalidArgumentException;
-
-use function in_array;
-
-/**
- * @internal this is a support class of this library, and should NOT be used outside of it
```
Ocramius, line 12: Good riddance!

Ocramius (changes requested):
Overall massive improvements, but I would suggest rolling back all changes in which the core API is:

 * safe
 * pure
 * acting exactly like PSL

There is tremendous value in replacing `Safe\` with `Psl\`, but mostly risk and overhead when instead we replace pure core constructs with `Psl\`.

The reason there is risk is that many of these core functions are correctly inferred by psalm, as well as tooling like `infection/infection`, and have near-zero overhead, while `Psl\` constructs are not the same there.

Functions that IMO should be rolled back:

 * `array_intersect_*`
 * `array_values`
 * `array_filter`
 * `array_map`
 * `strtolower`
 * `array_keys`
 * `array_combine`
 * `implode`
 * ... perhaps more? ...

As for `Shell\execute()` calls, before we move on and phase out `symfony/process`, what will happen if a security issue occurs on `azjezz/psl` on old branches? Will it be fixed in all past branches?

src/DetectChanges/BCBreak/ClassBased/MethodChanged.php
azjezz: consistency, that's the only reason, as well as guaranteed consistent behavior when it comes to type erorrs ( this https://wiki.php.net/rfc/consistent_type_errors, but for PHP 7.4 )

src/CompareClasses.php
azjezz: https://3v4l.org/KOgP7

azjezz: consistency ( in errors especially ) was the main reason for these functions, but at some point i forgot why i added them and made them "aliases" for the sake of performance ( see https://github.com/azjezz/psl/pull/169/files ), i think the middle ground is to wrap the closures again before passing them to builtin functions ( will release this as a bug fix in PSL ), see https://3v4l.org/TPJYP

azjezz:
> acting exactly like PSL
> mostly risk and overhead when instead we replace pure core constructs with Psl\.

The difference as mentioned is consistency in error handling  ( as in consistency between your code, and the standard library ), parameter order, and naming.

>  As for Shell\execute() calls, before we move on and phase out symfony/process, what will happen if a security issue occurs on azjezz/psl on old branches? Will it be fixed in all past branches?

Yes, security fixes will apply to past branches, as well as bug fixes.

I will probably add a document later to PSL explaining which versions will receive bug fixes/security patches and for how long, there's already a security policy in place to report vulnerabilities.

src/CompareClasses.php
Ocramius: Wait what, `array_*` functions run with non-strict types all the times? O_O

azjezz: you have no idea about the amount of inconsistency i figured out while working on PSL.

src/DetectChanges/BCBreak/ClassBased/AncestorRemoved.php
Ocramius: As per discussion with https://github.com/Roave/BackwardCompatibilityCheck/pull/306#discussion_r635004153, will roll with the `Psl\` version for now, as it is seems like I underestimated just how bad internal PHP functions are :|

src/CompareClasses.php
Ocramius: Cool, let's keep the PSL version then :+1:

bin/roave-backward-compatibility-check.php
Ocramius: Unrelated to patch - will keep

src/CompareClasses.php
azjezz: i will release a bug fix today in PSL ( https://github.com/azjezz/psl/issues/194 ), then i will finish this one :)

Ocramius:
@azjezz I marked all feedback related to restoring internal functions as "resolved", since you've convinced me to fully buy into `Psl\` there :-)

3 comments remain, then it's mergeable, IMO.

Ocramius:
BTW, wanted to note that `Psl\` is not really understood by `infection/infection`, and the number of mutants changed quite a lot (both a positive and a negative):

```diff
- 372 mutations were generated:
+ 263 mutations were generated:
-     371 mutants were killed
+     262 mutants were killed
       0 mutants were not covered by tests
       0 covered mutants were not detected
       1 errors were encountered
       0 time outs were encountered
```

azjezz:
hmm, I can probably create an adapter under @php-standard-library org, i didn't get to play with infection much since https://github.com/infection/infection/pull/1483 is still WIP :(

Ocramius:
@azjezz don't think it's really needed, just wanted to highlight that many mutations are possible because the libraries "know" how to deal with internal symbols, while aren't able to understand custom symbols.

This is normal, and it's just part of the considerations.

azjezz:
done :)

MGatner:
I'm just an observer here, but this is a super impressive PR so I'm compelled to give props!

PR #324, opened 2021-11-06
PHP 8.0 and PHP 8.1 support, BetterReflection 5.x upgrade

Ocramius:
This patch brings in PHP 8.0 and 8.1 support through the massive work done by @kukulich on
`roave/better-reflection` over the past few months.

The upgrade is mostly smooth, but we have some rough edges with:

 * [x] `TypeIsCovariant`, which needs `ReflectionType`s as well as `Reflector`s
 * [x] `TypeIsContravariant`, which needs `ReflectionType`s as well as `Reflector`s
 * [x] `ReturnTypeCovarianceChanged` uses hacky reflection to get a `Reflector` from a `ReflectionFunctionAbstract` (internal state!) - needs https://github.com/Roave/BetterReflection/pull/836
 * [x] `ParameterTypeContravarianceChanged` uses hacky reflection to get a `Reflector` from a `ReflectionFunctionAbstract` (internal state!) - needs https://github.com/Roave/BetterReflection/pull/836
 * [x] we miss tests for covariance/contravariance with `ReflectionUnionType` and `ReflectionIntersectionType` (`infection/infection` will highlight that)
 * [x] https://github.com/Roave/BetterReflection/pull/902
 * [x] tests for covariance/contravariance of nullable `ReflectionIntersectionType` (needs https://github.com/Roave/BetterReflection/pull/902 )

Fixes #283

PR #340, opened 2021-11-27
Add testsand implementation for property types changing from docblock-only to native type

tomasnorre:
As see yesterday, native property types aren't supported currently. 
https://twitter.com/tomasnorre/status/1464301968264994816

I have now added tests that proves this. I'm still trying to get my head around how to implement the change, I'm not really strong at AST, even though I have written some Rector rules. 

Any pointers on how to approach this would be helpful, thanks.

### DIFF
--- a/test/unit/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChangedTest.php
+++ b/test/unit/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChangedTest.php
@@ -222,7 +222,7 @@ PHP
             'propertyTypeBeingDuplicatedAreNotBcBreaks'                         => [],
             'propertyWithComplexDocblockThatCannotBeParsed'                     => [],
             'propertyWithDocblockTypeHintChangeToNativeTypeHint'                => [],
-            'propertyWithDocblockTypeHintChangeToNativeTypeHintAndTypeChange'   => ['[BC] CHANGED: Type documentation for property TheClass#propertyWithDocblockTypeHintChangeToNativeTypeHintAndTypeChange changed from int to float'],
+            'propertyWithDocblockTypeHintChangeToNativeTypeHintAndTypeChange'   => ['[BC] CHANGED: Type documentation for property TheClass#$propertyWithDocblockTypeHintChangeToNativeTypeHintAndTypeChange changed from int to float'],
         ];
 
         return array_combine(
--- a/test/unit/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChangedTest.php
+++ b/test/unit/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChangedTest.php
@@ -119,7 +119,10 @@ class TheClass {
      */
     public $propertyWithDocblockTypeHintChangeToNativeTypeHint;
     
-    public int $propertyWithNativeTypeHintChanged;
+    /**
+     * @var int
+     */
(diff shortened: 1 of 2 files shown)


test/unit/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChangedTest.php
```diff
@@ -113,6 +113,13 @@ class TheClass {
      * @var GenericType<T1, T2>
      */ 
     public $propertyWithComplexDocblockThatCannotBeParsed;
+    
+    /**
+     * @var int
+     */
+    public $propertyWithDocblockTypeHintChangeToNativeTypeHint;
+    
+    public int $propertyWithNativeTypeHintChanged;
```
Ocramius: I don't think changes in native types should be handled here: new checker would be better here

test/unit/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChangedTest.php
```diff
@@ -113,6 +113,13 @@ class TheClass {
      * @var GenericType<T1, T2>
      */ 
     public $propertyWithComplexDocblockThatCannotBeParsed;
+    
+    /**
+     * @var int
```
Ocramius, line 118: This looks like a good/valid scenario: type changed when moving from docblock to native.

Do we have a scenario where the same happens, but the type didn't change?

Ocramius (changes requested):

test/unit/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChangedTest.php
tomasnorre: The type didn't change in this example, or am I misunderstanding you.

Ocramius: Sorry, yes, I inverted it while writing the comment 🤦

What I basically mean is that we need both a positive and negative scenario for docblock -> native changes.

tomasnorre: Ok. Will remove the one that you think don't belong here, and add the second instead.

tomasnorre:
I'm not sure about the implementation, but would love some feedback on it.

Ocramius (approved):
LGTM :+1:

Ocramius:
Thanks @tomasnorre!

tomasnorre:
I don't know if this is a new PR, but I saw this, this morning, after using the 6.0.x-dev version in my CI. 

```
AOE\Crawler\Controller\CrawlerController#$queueRepository changed from \AOE\Crawler\Domain\Repository\QueueRepository to AOE\Crawler\Domain\Repository\QueueRepository
``` 

The diff is the leading slash. Is stripping the leading slash before compare a valid fix? 
As I'm using `use` statements and have short names, there will not be leading slashes in the native types. 

The rest of the cases I had is fixes with the current version of 6.0.x

Ocramius:
Would always start from a test case, then figure it out from there.

tomasnorre:
I'll create a new PR with a Test that fails. Probably not today, though.

Ocramius:
Important, before you consider adding more work to this: https://github.com/Roave/BackwardCompatibilityCheck/pull/324/commits/f8c25b356ab752fc962faceae1f72e0566604480

I had to butcher your previous work to move forward :-(

PR #410, opened 2022-01-15
Replace file path when it is encountered in string constants

### LINKED ISSUE #400: False positive result when __DIR__ is used
Example:

```php
final class EngineLoaderFactory implements EngineLoaderFactoryInterface
{
    public const DATA_PATH = __DIR__ . '/../../data/engines';
}
```

Expected Result: no error

Result:
```shell
# Changed
 - [BC] Value of constant BrowserDetector\Loader\EngineLoaderFactory::DATA_PATH changed from '/tmp/api-compare-ed688120622a186b52c6e1b33a3065507a5f04e961daa0714f8ee/src/Loader/../../data/engines' to '/tmp/api-compare-b5c256f5ff06bbf49461563348993ebb614d660961daa073ba0ab/src/Loader/../../data/engines'
```

asgrim:
Fixes #400

### DIFF
--- a/src/DetectChanges/BCBreak/ClassConstantBased/ClassConstantValueChanged.php
+++ b/src/DetectChanges/BCBreak/ClassConstantBased/ClassConstantValueChanged.php
@@ -17,7 +17,7 @@ use function var_export;
 
 final class ClassConstantValueChanged implements ClassConstantBased
 {
-    private const MAGIC_DIR_OR_FILE_VALUE = '__DIR_OR_FILE__';
+    private const MAGIC_DIR_OR_FILE_VALUE = '__DIR_HAS_BEEN_REPLACED__';
 
     public function __invoke(ReflectionClassConstant $fromConstant, ReflectionClassConstant $toConstant): Changes
     {
--- a/src/DetectChanges/BCBreak/ClassConstantBased/ClassConstantValueChanged.php
+++ b/src/DetectChanges/BCBreak/ClassConstantBased/ClassConstantValueChanged.php
@@ -9,10 +9,16 @@ use Roave\BackwardCompatibility\Change;
 use Roave\BackwardCompatibility\Changes;
 use Roave\BetterReflection\Reflection\ReflectionClassConstant;
 
+use function dirname;
+use function is_string;
+use function realpath;
+use function str_replace;
 use function var_export;
 
 final class ClassConstantValueChanged implements ClassConstantBased
 {
+    private const MAGIC_DIR_OR_FILE_VALUE = '__DIR_OR_FILE__';
+
     public function __invoke(ReflectionClassConstant $fromConstant, ReflectionClassConstant $toConstant): Changes
     {
         if ($fromConstant->isPrivate()) {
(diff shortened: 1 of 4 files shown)


src/DetectChanges/BCBreak/ClassConstantBased/ClassConstantValueChanged.php
```diff
@@ -39,4 +45,20 @@ public function __invoke(ReflectionClassConstant $fromConstant, ReflectionClassC
...
+        if (! is_string($value)) {
+            return $value;
+        }
+
+        $filePath = $constant->getDeclaringClass()->getFileName();
+        if (! $filePath) {
+            return $value;
+        }
+
+        return str_replace(dirname(realpath($filePath)), self::MAGIC_DIR_OR_FILE_VALUE, $value);
```
asgrim, line 62: I'm not 100% convinced this is "the best way":tm: but it was just an idea I put together. Thoughts welcome!

Ocramius: Indeed, looks iffy

Ocramius (changes requested):
After thinking further about it, I think the correct way would be to change the path detected for source locators, which would fix a whole category of issues related to this.

I don't know yet how to approach this, but shadowing the `/tmp` location completely seems like a better approach.

Ocramius:
One approach I thought of (not sure if viable), would be to replace `LocatedSource` paths via decoration:

https://github.com/Roave/BackwardCompatibilityCheck/blob/5d242ba3defae9e1768ebb10c337abf8091f52bf/src/SourceLocator/StaticClassMapSourceLocator.php#L55

We could produce a `VirtualPathLocatedSource` or such, where `LocatedSource`'s path getter is decorated to remove the TMP path section 🤔

Ocramius:
I've potentially identified a way of making this work in a relatively hacky, but clean way.

The way we find symbols via source locators is via `Roave\BetterReflection\SourceLocator\Ast\Locator`.

While `Roave\BetterReflection\SourceLocator\Ast\Locator` is marked `@internal`, it is a core component to source locators, and its job is to convert string sources to reflection symbols.

We could **decorate** the `Locator`, and make sure that any `LocatedSource` instance passed to it is decorated, so that `LocatedSource#getFileName()` is stripped of any temporary source path.

That would make `__DIR__` and `__FILE__` work as expected again, I think.

asgrim:
@Ocramius that sounds like a plausible approach; will try and look into this when I can

Ocramius:
I sent up a patch with my proposed approach @ https://github.com/Roave/BackwardCompatibilityCheck/pull/476

It will need patching `roave/better-reflection` to get rid of `realpath()` and `dirname()` from its code, but unsure if that's feasible at all.

Ocramius:
Closing this one: approach in #476 works fine, and is a bit better than operating on the constant expressions, as those can really occur almost anywhere in the system, so relying on the abstractions by `roave/better-reflection` is indeed to be prioritized.

PR #411, opened 2022-01-15
Added support for `@no-named-arguments` on the class phpdoc block

### LINKED ISSUE #409: Support @no-named-arguments when applied to whole class
https://github.com/Roave/BackwardCompatibilityCheck/blob/316285bd5d3788d7fd1c28d99b76dcac64e3074a/src/DetectChanges/BCBreak/FunctionBased/ParameterNameChanged.php#L114-L117

The current implementation only recognises if `@no-named-arguments` on the method/function itself.

Apparently, PHPUnit uses this on the class-level too, which is a reasonable assumption that is missed out here:

```
/**
 * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
 */
abstract class TestCase ...
```

asgrim:
Fixes #409


Ocramius (approved):
LGTM 👍

Ocramius:
Thanks @asgrim! 💪

PR #476, opened 2022-03-08
Normalize `__DIR__` and `__FILE__` by changing the base path of `LocatedSource` instances, prevent false-positive BC breaks on those magic constants

### LINKED ISSUE #400: False positive result when __DIR__ is used

Ocramius:
This fails due to `CompileNodeToValue()` internally using `realpath()` and `dirname()` operations, which both depend
on the current filesystem: as it currently stands, this requires an improvement of `roave/better-reflection` first.

Also, `ReplaceSourcePathOfLocatedSources` and `LocatedSourceWithStrippedSourcesDirectory` are extending `@internal`
symbols, so we may want to first make these symbols public in `roave/better-reflection`, to avoid the BC boundary to
become too relaxed, and therefore causing crashes if we (rightfully) decide to change `roave/better-reflection` internals.

Fixes #400
Fixes #410

## TODO

 * [x] https://github.com/Roave/BetterReflection/issues/1043

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "composer/composer": "^2.2.9",
         "nikolaposa/version": "^4.1.0",
         "ocramius/package-versions": "^2.5.1",
-        "roave/better-reflection": "^5.2.0",
+        "roave/better-reflection": "^5.3.0",
         "symfony/console": "^5.4.5"
     },
     "license": "MIT",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "ae7d8dc1e7f6cc719393a69033a395d8",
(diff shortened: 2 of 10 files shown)


Ocramius:
@asgrim this is (IMO) ready for review

asgrim (approved):
Yeah this is much cleaner - looks like it was a bit of a journey to get here though heh. Nice one :+1:

Ocramius:
:ship: then, thanks for the review, @asgrim!

PR #522, opened 2022-05-22
Updated PHAR release script

### LINKED ISSUE #374: Phar release is not created
As per:

https://twitter.com/s_bergmann/status/1475014835129757699

Looks like the "release-phar" action did not run (can't verify, I'm on phone)

It should be triggered by creating the release. However, my guess is that because we're using laminas/automatic-releases, downstream events are not triggered due to the GitHub access token being used. It might be possible to switch this out to use the personal access token for the organisation, which would cause downstream events to trigger actions in turn.

https://docs.github.com/en/actions/reference/authentication-in-a-workflow#using-the-github_token-in-a-workflow

> When you use the repository's `GITHUB_TOKEN` to perform tasks on behalf of the GitHub Actions app, events triggered by the `GITHUB_TOKEN` will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's `GITHUB_TOKEN`, a new workflow will not run even when the repository contains a workflow configured to run when `push` events occur.

asgrim:
Fixes #374

NOTE - I have not tested this end to end - not entirely sure how to set up GPG to use the private key without nuking my own private key.

The phar builds (example attached, zipped so that GH accepts it) but I'm not sure how to test the signing

[roave-backward-compatibility-check.zip](https://github.com/Roave/BackwardCompatibilityCheck/files/8749218/roave-backward-compatibility-check.zip)
.

### DIFF
--- a/.github/workflows/release-phar.yml
+++ b/.github/workflows/release-phar.yml
@@ -4,61 +4,47 @@ name: "Release PHAR"
 
 on:
   release:
-    types:
-      - published
+    types: [published]
 
 jobs:
-  release-phar:
-    name: "Release PHAR"
+  build:
+    name: "Building and upload PHAR artifact"
 
-    runs-on: ${{ matrix.operating-system }}
-
-    strategy:
-      matrix:
-        dependencies:
-          - "locked"
-        php-version:
-          - "7.4"
-        operating-system:
-          - "ubuntu-latest"
+    runs-on: "ubuntu-latest"
 
     steps:
       - name: "Checkout"
-        uses: "actions/checkout@v3"
+        uses: "actions/checkout@v2"
 
       - name: "Install PHP"
         uses: "shivammathur/setup-php@v2"
         with:
-          coverage: "pcov"
-          php-version: "${{ matrix.php-version }}"
-          ini-values: memory_limit=-1
+          coverage: "none"
+          php-version: "8.0"
+          ini-values: memory_limit=-1, phar.readonly=0
 
-      - name: "Cache dependencies"
-        uses: "actions/cache@v3"
-        with:
-          path: |
-            ~/.composer/cache
-            vendor
-          key: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}"
-          restore-keys: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}"
+      - name: "Import GPG Key"
+        run: echo "$PRIVATE_KEY" | gpg --import
+        env:
+          PRIVATE_KEY: ${{ secrets.SIGNING_SECRET_KEY }}
 
-      - name: "Install lowest dependencies"
-        if: ${{ matrix.dependencies == 'lowest' }}
-        run: "composer update --prefer-lowest --no-interaction --no-progress --no-suggest"
-
-      - name: "Install highest dependencies"
-        if: ${{ matrix.dependencies == 'highest' }}
-        run: "composer update --no-interaction --no-progress --no-suggest"
-
-      - name: "Install locked dependencies"
-        if: ${{ matrix.dependencies == 'locked' }}
+      - name: "Install dependencies for build environment"
         run: "composer install --no-interaction --no-progress --no-suggest"
 
-      - name: "build-phar.sh"
-        run: "build-phar.sh"
+      - name: "Build and sign phar file via phing"
+        run: "vendor/bin/phing phar-build phar-sign"
 
-      - name: "Upload PHAR"
-        uses: fnkr/github-action-ghr@v1
-        env:
-          GHR_PATH: dist/
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: "Upload phar file artifact"
+        uses: actions/upload-artifact@v2
+        with:
+          name: composer-require-checker.phar
+          path: |
+            build/composer-require-checker.phar
+            build/composer-require-checker.phar.asc
+      - name: Upload PHAR to release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: build/composer-require-checker.phar*
+          file_glob: true
+          tag: ${{ github.ref }}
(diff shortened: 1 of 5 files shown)


.github/workflows/release-phar.yml
```diff
@@ -4,61 +4,47 @@ name: "Release PHAR"
...
 
-      - name: "Install locked dependencies"
-        if: ${{ matrix.dependencies == 'locked' }}
+      - name: "Install dependencies for build environment"
         run: "composer install --no-interaction --no-progress --no-suggest"
 
-      - name: "build-phar.sh"
-        run: "build-phar.sh"
+      - name: "Build and sign phar file via phing"
+        run: "vendor/bin/phing phar-build phar-sign"
```
asgrim, line 35: I was able to run `phar-build` locally to produce a phar (attached to the main PR description), but I am not sure how to test the signing, and uploading the PHAR to the release.

.github/workflows/release-phar.yml
```diff
@@ -4,61 +4,47 @@ name: "Release PHAR"
...
-        php-version:
-          - "7.4"
-        operating-system:
-          - "ubuntu-latest"
+    runs-on: "ubuntu-latest"
 
     steps:
       - name: "Checkout"
-        uses: "actions/checkout@v3"
+        uses: "actions/checkout@v2"
```
Ocramius, line 17: To be reverted

.github/workflows/release-phar.yml
Ocramius: Need to do a release 😬

build.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- ============================================================== -->
+<!-- Phing Build instructions                                       -->
+<!-- http://www.phing.info/                                         -->
+<!-- http://www.phing.info/get/phing-latest.phar                    -->
+<!-- php -d phar.readonly=Off phing-latest.phar phar-build-release  -->
+<!-- ============================================================== -->
+<project name="roave-backward-compatibility-check" default="phar-build">
```
Ocramius, line 8: Why the fresh hell do we need phing, and why does it still exist? 😱

composer.json
```diff
@@ -24,6 +24,7 @@
     ],
     "require-dev": {
         "doctrine/coding-standard": "^9.0.0",
+        "phing/phing": "^2.17",
```
Ocramius, line 27: Nopenopenope 🤣

build.xml
```diff
@@ -0,0 +1,61 @@
...
+    </target>
+
+    <!-- ============================================  -->
+    <!-- Target: phar-build                            -->
+    <!-- ============================================  -->
+    <target name="phar-build" depends="phar-prepare-dependencies">
+        <!--create the package-->
+        <mkdir dir="${build-dir}" />
+        <php expression="file_put_contents('bin/clistub.php', '#!/usr/bin/env php' . chr(10) . Phar::createDefaultStub('bin/roave-backward-compatibility-check.php'))"/>
+        <pharpackage basedir="./"
```
Ocramius, line 27: We really shouldn't drag in phing for just this 😬

Ocramius (changes requested):

build.xml
asgrim: This is just based off the suggested approach in maglnet/ComposerRequireChecker - which uses Phing's PHAR packager to make this heh

Ocramius: Yup - I think we just need to extract this specific step to be done via bash, and then we are good 🤔

Ocramius: Seems like it isn't too much Voodoo: https://github.com/phingofficial/phing/blob/d98239f270e86c6a886eb93893d5ddd17c9d71bd/src/Phing/Task/System/PharPackageTask.php

Only dependency becomes ext-phar in require-dev

sebastianbergmann: Sorry for bothering: any update on this? Thanks!

Ocramius: /cc @asgrim ofr this: I completely dropped the ball on PHARs on my end.

asgrim: I'm not sure how else to do this? The phing approach gets it done, but you seem hesitant to bring in the dep... Don't know about other approaches at this time.

Ocramius: Hmm, yeah, looking at https://github.com/maglnet/ComposerRequireChecker/blob/fb2a69aa2b7307541233536f179275e99451b339/.github/workflows/phar-creation.yml#L32, there's no real way around it.

We could have a `tools/phing/composer.json`, but let's roll with the current solution for now, if you are happy with it.

I can extract it from here later, if it becomes a problem: would really rather not have phing in my big list of problems, but it is what it is :D

asgrim: Not sure if @theseer might be working a tool: https://phpc.social/@theseer/109296480712703364

Ocramius: Let's just go with what works now and unblocks @sebastianbergmann, and then we suffer later in trying to extract it from our pipeline again :P

asgrim: OK, I'll take a look at this again and refresh myself, haven't looked at it for a while and probably needs a rebase etc :grin:

theseer: > Not sure if @theseer might be working a tool: [phpc.social/@theseer/109296480712703364](https://phpc.social/@theseer/109296480712703364)

This is work in progress and in early design phase on my side.
Given I'm currently quite busy, this is likely not to materialize very soon :-(

mfn:
Isn't this obsolete since the PHAR build process has been removed? -> https://github.com/Roave/BackwardCompatibilityCheck/pull/770

Ocramius:
Yup, thanks for noticing, @mfn

mfn:
I actually was searching for the PHAR for CI integration but first found this issue and was in hopes until I found the other 😭

Ocramius:
@mfn the correct way to do this, in my (long standing) opinion:

* make a `tools/` directory
* add a `composer.json` and `composer.lock` to just that directory 
* run the tool from `./tools/vendor/bin/roave-backward-compatibility-check`

mfn:
To me this was totally not known. And it makes a lot of sense!

I've successfully applied this strategy to public and private codebases in the meantime, it's extremely liberating from dev dependency hell.

I feel like the README could promote this more 👍🏼

Ocramius:
@mfn send a patch/proposal: happy to merge it :)

PR #643, opened 2022-10-10
Support PHP 8.2, drop support for PHP 8.0, upgrade to `azjezz/psl:^2`

Ocramius:


Ocramius:
As discussed with @azjezz, will need https://github.com/azjezz/psl/pull/363 to be released in upstream before this can be merged.

Issue #653, opened 2022-10-11
Using this package in Bitbucket Pipelines


mad-briller:
First things first, this package is a genius idea, thank you for the work on it.

However we are trying to integrate it with Bitbucket pipelines and insights, but the output formats available are awkward for us to consume in this context.

`github-actions` is the most machine-readable but that requires us to implement a custom parser for the format to transform into the format required by bitbucket (unless there is a php implementation to parse this already? i had no luck finding one)

would Roave be open to a PR that implemented a `json` formatter type that would make it simpler for us to consume?

thanks for your time

Ocramius:
Sure thing, but if we provide a JSON formatter, it must come with an associated JSON Schema

Ocramius:
BTW, you can check for more standard formats, like XUnit and similar: some of them work well with both Bitbucket and GitLab

mad-briller:
Bitbucket accepts those as test-reports but won't use them for the purpose of annotating pull requests unfortunately; annotating pull requests in the bitbucket world requires making api requests, which is very awkward. 

However it still stands to reason that we could use a more standard format, i will look into JUnit as it's one i'm more familiar with. Don't know why i didn't think of that first, thanks!

Ocramius:
Yeah, even if the reporting requires making API calls, tooling for that probably already exists, given a good input format.

mad-briller:
thanks for your help with this @Ocramius

Ocramius:
Thank you for implementing this with a clean design! 👍

PR #654, opened 2022-10-11
Added junit formatter.

mad-briller:
refs: https://github.com/Roave/BackwardCompatibilityCheck/issues/653

### DIFF
--- a/src/Command/AssertBackwardsCompatible.php
+++ b/src/Command/AssertBackwardsCompatible.php
@@ -13,6 +13,7 @@ use Roave\BackwardCompatibility\Changes;
 use Roave\BackwardCompatibility\CompareApi;
 use Roave\BackwardCompatibility\Factory\ComposerInstallationReflectorFactory;
 use Roave\BackwardCompatibility\Formatter\GithubActionsFormatter;
+use Roave\BackwardCompatibility\Formatter\JunitFormatter;
 use Roave\BackwardCompatibility\Formatter\MarkdownPipedToSymfonyConsoleFormatter;
 use Roave\BackwardCompatibility\Formatter\SymfonyConsoleTextFormatter;
 use Roave\BackwardCompatibility\Git\CheckedOutRepository;
@@ -69,7 +70,7 @@ final class AssertBackwardsCompatible extends Command
                 'format',
                 null,
                 InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
-                'Currently supports "console", "markdown" or "github-actions"',
+                'Currently supports "console", "markdown", "github-actions" or "junit"',
                 ['console'],
             )
             ->addOption(
@@ -82,7 +83,7 @@ final class AssertBackwardsCompatible extends Command
                 <<<'USAGE'
 
 
-Without arguments, this command will attempt to detect the 
+Without arguments, this command will attempt to detect the
 latest stable git tag ("release", according to this tool)
 of the repository in your CWD (current working directory),
 and will use it as baseline for the defined API.
@@ -152,6 +153,7 @@ USAGE,
                 'console'        => new SymfonyConsoleTextFormatter($stdErr),
                 'markdown'       => new MarkdownPipedToSymfonyConsoleFormatter($output),
                 'github-actions' => new GithubActionsFormatter($output, $toPath),
+                'junit'          => new JunitFormatter($output, $toPath),
             ];
 
             foreach (
@@ -159,6 +161,7 @@ USAGE,
                     Type\literal_scalar('console'),
                     Type\literal_scalar('markdown'),
                     Type\literal_scalar('github-actions'),
+                    Type\literal_scalar('junit'),
                 ))->coerce((array) $input->getOption('format')) as $format
             ) {
                 $formatters[$format]->write($changes);
--- a/src/Command/AssertBackwardsCompatible.php
+++ b/src/Command/AssertBackwardsCompatible.php
@@ -72,6 +72,7 @@ final class AssertBackwardsCompatible extends Command
                 InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
                 'Currently supports "console", "markdown", "github-actions" or "junit"',
                 ['console'],
+                ['console', 'markdown', 'github-actions', 'junit'],
             )
             ->addOption(
                 'install-development-dependencies',
--- a/src/Formatter/JunitFormatter.php
+++ b/src/Formatter/JunitFormatter.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BackwardCompatibility\Formatter;
+
+use Psl\Str;
+use Roave\BackwardCompatibility\Changes;
+use Roave\BackwardCompatibility\Git\CheckedOutRepository;
+use Symfony\Component\Console\Output\OutputInterface;
+
+use function count;
+use function htmlspecialchars;
+use function implode;
+use function sprintf;
+use function trim;
+
+use const ENT_COMPAT;
+use const ENT_XML1;
+
+class JunitFormatter implements OutputFormatter
+{
+    public function __construct(
+        private OutputInterface $output,
+        private CheckedOutRepository $basePath,
+    ) {
+    }
+
+    public function write(Changes $changes): void
+    {
+        $basePath = $this->basePath->__toString() . '/';
+
+        $changeCount = count($changes);
+
+        $testcases = [];
+
+        foreach ($changes as $change) {
+            $filename = $change->file === null ? null : Str\replace($change->file, $basePath, '');
+
+            $name = implode(':', [
+                $this->escape($filename ?? ''),
+                $this->escape((string) ($change->line ?? '')),
+                $this->escape((string) ($change->column ?? '')),
+            ]);
+
+            $testcases[] = sprintf(
+                '  <testcase name="%s"><failure type="error" message="%s"/></testcase>',
+                $this->escape($name),
+                $this->escape(trim($change->__toString())),
+            );
+        }
+
+        $result = implode("\n", [
+            '<?xml version="1.0" encoding="UTF-8"?>',
+            sprintf(
+                '<testsuite name="roave/backward-compatibility-check" tests="%d" failures="%d" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/junit-team/junit5/r5.9.1/platform-tests/src/test/resources/jenkins-junit.xsd">',
+                $changeCount,
+                $changeCount,
+            ),
+            ...$testcases,
+            '</testsuite>',
+        ]);
+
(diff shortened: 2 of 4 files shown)


src/Command/AssertBackwardsCompatible.php
```diff
@@ -69,7 +70,7 @@ protected function configure(): void
                 'format',
                 null,
                 InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
-                'Currently supports "console", "markdown" or "github-actions"',
+                'Currently supports "console", "markdown", "github-actions" or "junit"',
```
Ocramius, line 73: I wonder if it's possible to give the options as a `list<string>` to `addOption()`, so it prints the possible values (and does the validation) later?

src/Formatter/JunitFormatter.php
```diff
@@ -0,0 +1,71 @@
...
+use function count;
+use function htmlspecialchars;
+use function implode;
+use function sprintf;
+use function trim;
+
+use const ENT_COMPAT;
+use const ENT_XML1;
+
+class JunitFormatter implements OutputFormatter
```
Ocramius: `final`

src/Formatter/JunitFormatter.php
```diff
@@ -0,0 +1,71 @@
...
+{
+    public function __construct(
+        private OutputInterface $output,
```
Ocramius: `readonly` should be added to these

src/Formatter/JunitFormatter.php
```diff
@@ -0,0 +1,71 @@
...
+            $filename = $change->file === null ? null : Str\replace($change->file, $basePath, '');
+
+            $name = implode(':', [
+                $this->escape($filename ?? ''),
+                $this->escape((string) ($change->line ?? '')),
+                $this->escape((string) ($change->column ?? '')),
+            ]);
+
+            $testcases[] = sprintf(
+                '  <testcase name="%s"><failure type="error" message="%s"/></testcase>',
```
Ocramius, line 57: If we use an XML format, we should probably use a DOM builder for this, no? :thinking: 

String concatenation doesn't seem very healthy, although I see that you escape things accordingly

src/Formatter/JunitFormatter.php
```diff
@@ -0,0 +1,71 @@
...
+                $changeCount,
+            ),
+            ...$testcases,
+            '</testsuite>',
+        ]);
+
+        $this->output->writeLn($result);
+    }
+
+    private function escape(string $value): string
```
Ocramius: Should probably rename this to `escapeXmlAttribute()`

src/Formatter/JunitFormatter.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        return htmlspecialchars($value, ENT_XML1 | ENT_COMPAT, 'UTF-8');
```
Ocramius, line 68: Are quotes considered in this escaping?

src/Formatter/JunitFormatter.php
```diff
@@ -0,0 +1,71 @@
...
+        foreach ($changes as $change) {
+            $filename = $change->file === null ? null : Str\replace($change->file, $basePath, '');
+
+            $name = implode(':', [
+                $this->escape($filename ?? ''),
+                $this->escape((string) ($change->line ?? '')),
+                $this->escape((string) ($change->column ?? '')),
+            ]);
+
+            $testcases[] = sprintf(
```
Ocramius: Note: the output should probably be streamed.

A crash mid-way would make this reporter really annoying to work with :thinking:

test/unit/Formatter/junit.xsd
```diff
@@ -0,0 +1,118 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR   *
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,     *
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE  *
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER       *
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,*
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN    *
+ * THE SOFTWARE.                                                                *
+ ********************************************************************************
+ -->
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
```
Ocramius, line 29: Thanks for including an XSD! :+1:

test/unit/Formatter/JunitFormatterTest.php
```diff
@@ -0,0 +1,90 @@
...
+        ));
+
+        Filesystem\delete_directory($temporaryLocation, true);
+
+        $fetchedOutput = $output->fetch();
+
+        self::assertSame(
+            <<<'OUTPUT'
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuite name="roave/backward-compatibility-check" tests="7" failures="7" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/junit-team/junit5/r5.9.1/platform-tests/src/test/resources/jenkins-junit.xsd">
```
Ocramius: Let's use an absolute reference, rather than a tag? That would prevent the tag from moving, as well as XXE/XEE in the test suite being reduced

Ocramius (changes requested):

src/Formatter/JunitFormatter.php
mad-briller: i considered using a DOMDocument to build it but that would require the `dom` extension which i wasn't sure on the stance that you'd take

i can rework it to use it and have an `extension_loaded` check at the start of the formatter and throw if it's not present?

src/Formatter/JunitFormatter.php
mad-briller: good question, i'll investigate tomorrow

src/Formatter/JunitFormatter.php
mad-briller: would streaming an xml document be useful? would interacting with a partial xml document ever be done? most xml parsers would fail i imagine

it wouldn't be too difficult to output line by line however, but if i work with `DOMDocument` instead the ability to stream it would be removed... i think?

src/Formatter/JunitFormatter.php
Ocramius: Given that the escaping is done well, I think we can live with the string concatenation approach: we just need to make sure we note that in a comment (on the class) for posterity :+1:

src/Formatter/JunitFormatter.php
Ocramius: Streaming useful when this crashes mid-way.

But then... I noticed that we need to `count()` the results anyway, which defeats the streaming, I guess? :thinking: 

The problem here is that analysis can take between seconds and minutes, and no output for multiple minutes can be problematic for some CI environments.

test/unit/Formatter/JunitFormatterTest.php
mad-briller: i guess you are referring to the `r5.9.1` part of the url here, but sorry, i'm not sure what you mean by an absolute reference?

referencing any external resource will always poise the opportunity for XXE i suppose

Ocramius: Tags can mutate: much harder to do so with SHA1s :D 

Here's the absolute URI for that: https://raw.githubusercontent.com/junit-team/junit5/732a5400f80c8f446daa8b43eaa4b41b3da929be/platform-tests/src/test/resources/jenkins-junit.xsd

src/Formatter/JunitFormatter.php
mad-briller: i guess that's why the github-actions output is the way that it is; it allows for partial and streaming output in a way that json / xml does not allow, as each line stands on its own

judging by the OutputFormatter interface, all the results are gathered into `Changes` before being passed to the formatter it looks like? or am i wrong there and multiple calls to `write()` can be made in one analysis?

test/unit/Formatter/JunitFormatterTest.php
mad-briller: ahhh, the commit hash of course, thanks for that

src/Command/AssertBackwardsCompatible.php
mad-briller: the 6th argument to `addOption` `$suggestedValues` appears to take an array of values, but when used as such:

```php
            ->addOption(
                'format',
                null,
                InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
                'The output format to use.',
                ['console'],
                ['console', 'markdown', 'github-actions', 'junit'],
            )
```

the output looks like this: 

```
      --format[=FORMAT]                   The output format to use. [default: ["console"]] (multiple values allowed)
```

which unfortunately does not list the values

it appears as though the `$suggestedValues` parameter powers the autocompletion script made available in symfony console, and is not used to inform / validate the input beyond that

mad-briller: i will still add the 6th argument to enable the autocompletion however

src/Formatter/JunitFormatter.php
mad-briller: i've been thinking about this and the common use case of this format would be redirecting it to a file right? 

`php vendor/bin/backwards-compatibility-check --format=junit > junit.xml`
that means the CI would not see any output from the process anyway due to the redirection.

this is kind of what i meant by a partial xml document, if the output was streamed line by line to a file and the process crashed, the produced file would be incomplete and lacking the closing `</testsuite>` tag, and be invalid xml.


> judging by the OutputFormatter interface, all the results are gathered into Changes before being passed to the formatter it looks like? or am i wrong there and multiple calls to write() can be made in one analysis? 

upon looking into this it appears that all `Changes` are provided to the formatter in one go and it is not called multiple times, so i'm not sure there is a lot of value in streaming the output of the junit formatter?

however i might be missing something, so please let me know if i am.

Ocramius: Note: I use `| tee` to print even when redirecting 👍

src/Formatter/JunitFormatter.php
mad-briller: `ENT_XML1`: Handle code as XML 1
`ENT_COMPAT`: Will convert double-quotes and leave single-quotes alone.

quotes are considered in the escaping

src/Formatter/JunitFormatter.php
mad-briller: fair point, it's not much of a hassle to change either way so i'll get it done

Ocramius (approved):
Excellent! :ship: 

Thanks @mad-briller, well done!

Cutting a release now :)

PR #657, opened 2022-10-12
Always cleanup the temporary directory in github actions formatter test even if the assertion fails.

mad-briller:
Noticed while working on https://github.com/Roave/BackwardCompatibilityCheck/pull/654

if the assertion fails, it will throw an exception and as a result not clean up the temporary directory created in the test case, and potentially pollute other test runs.


Ocramius (approved):
Thanks for noticing this! :+1:

PR #668, opened 2022-10-21
[WIP] - Inform when there is no minor version

Ferror:
I had probably the edge-case when my library had no minor versions. Only preleases. I was getting information that there are no versions at all. Which is false. There are versions. I would like just to improve the DX and show the proper message :)
```
In CollectionIsEmpty.php line 13:
                                                         
  Cannot get the first Version from an empty collection  
                                                         
```

### DIFF
--- a/src/Command/AssertBackwardsCompatible.php
+++ b/src/Command/AssertBackwardsCompatible.php
@@ -21,8 +21,8 @@ use Roave\BackwardCompatibility\Git\GetVersionCollection;
 use Roave\BackwardCompatibility\Git\ParseRevision;
 use Roave\BackwardCompatibility\Git\PerformCheckoutOfRevision;
 use Roave\BackwardCompatibility\Git\PickVersionFromVersionCollection;
-use Roave\BackwardCompatibility\Git\Revision;
 use Roave\BackwardCompatibility\LocateDependencies\LocateDependencies;
+use Roave\BackwardCompatibility\VersionConstraint\StableVersionConstraint;
 use Roave\BetterReflection\SourceLocator\Type\AggregateSourceLocator;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Exception\InvalidArgumentException;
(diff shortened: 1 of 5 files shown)


src/Git/PickLastMinorVersionFromCollection.php
```diff
@@ -24,6 +24,10 @@ public function assert(Version $version): bool
                 return ! $version->isPreRelease();
             }
         });
+        
+        if ($stableVersions->isEmpty()) {
+            throw new \Exception('Your library does not have any minor version');
```
Ocramius: Is a crash really the best approach? Perhaps it should be "green" when no BC breaks are detected instead?

Ocramius (changes requested):

src/Git/PickLastMinorVersionFromCollection.php
Ferror: Thanks for the feedback. That's why I kinda put it as a draft with WIP. And yes when there is no stable version - there won't be any BC breaks. Note or Info message could be applied.

Stay tuned 😄

src/Command/AssertBackwardsCompatible.php
```diff
@@ -115,9 +115,27 @@ public function execute(InputInterface $input, OutputInterface $output): int
...
+            $versions = $this->getVersions->fromRepository($sourceRepo);
+
+            Psl\invariant(Iter\count($versions) >= 1, 'Could not detect any released versions for the given repository');
+
+            $stableVersions = $versions->matching(new StableVersionConstraint());
+
+            if ($stableVersions->isEmpty()) {
+                $stdErr->writeln(Str\format('Your library does not have any stable versions. Therefore cannot have BC breaks.'));
+
+                return Command::SUCCESS;
```
Ocramius, line 130: I just realized another problem, sorry @_@ 

Assume somebody incorrectly installs this tool with a CI setup that did **not** pull tags during clone operations (shallow clone: default on github), this leads to an exit code zero by default, hiding the underlying problem.

Perhaps an exception or non-zero exit code was really the best approach.

Ocramius (changes requested):

src/Command/AssertBackwardsCompatible.php
Ferror: Actually, maybe we can kill two birds with one stone. One is a misconfiguration which we could verify and the second is an exception :)

Ocramius: >  One is a misconfiguration which we could verify

How do we verify it, if no `git fetch` has occurred? :thinking:

Ferror: Isn't that...

No tags -> possible misconfiguration so we can throw an exception. Like. THERE ARE NO TAGS. WE NEED THEM (yes, we gonna scream 😄)

Then at some point, we can assume that there are at least some tags (we verified that by the previous action), but we want to check out the stable ones. There are tags, but none of them are stable. Return success.

What do you think?

Ocramius: > THERE ARE NO TAGS. WE NEED THEM (yes, we gonna scream smile)

Works for me - indeed the tool cannot do auto-`--from=` without tags.

 > Then at some point, we can assume that there are at least some tags (we verified that by the previous action), but we want to check out the stable ones. There are tags, but none of them are stable. Return success.

Good enough :-)

PR #700, opened 2022-11-25
Handle reflection error in retrieving default values

### LINKED ISSUE #698: Handle new initializer and enum intiliazing as default value more gracefully
Right now as soon as you initialize an object as default value.
```php
public function foo(Bar $bar = new Bar()): void {}
```

 you get an error like this
```
Error: Unable to compile initializer in method Patchlevel\EventSourcing\Projection\Projectionist::boot() in file /src/Projection/Projectionist.php (line 11)
```

Or using an enum value as default

```php
public function foo(Bar $bar = Bar::Baz): void {}
```
you will receive this:
```
Error: Could not locate constant Patchlevel\EventSourcing\Projection\ProjectorStatus::New while trying to evaluate constant expression in method Patchlevel\EventSourcing\Projection\ProjectorStore\ProjectorState::__construct() in file /src/Projection/ProjectorStore/ProjectorState.php (line 14)
```

This is because of `roave/better-reflection` is throwing an exception in these cases which is expected see https://github.com/Roave/BetterReflection/issues/1311 for more information.

In the linked issue we discussed to handle it here and catch the exception. My question here would be where should this happen? I looked into the code and couldnt really determine where the best place would be. Also, the thrown exception is marked `internal` at `roave/better-reflection` which lead me to the assumption to not catch it directly but instead `LogicException` or even `Throwable`.

DanielBadura:
fix #698 

This is the first PoC of handling the two errors mentioned in the linked issue. I added a TestCase for both errors. I'm not sure what exactly we should compare from the AST of the two parameters so for now only checking if the type is the same which already resolved the error case.


src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
```diff
@@ -38,8 +40,23 @@ public function __invoke(
...
+            try {
+                $defaultValueFrom = $parameter->getDefaultValue();
+                $defaultValueTo   = $toParametersWithDefaults[$parameterIndex]->getDefaultValue();
+            } catch (Throwable $throwable) {
+                $parameterDefaultExpression   = $parameter->getDefaultValueExpression();
+                $toParameterDefaultExpression = $toParametersWithDefaults[$parameterIndex]->getDefaultValueExpression();
+
+                if (
+                    $toParameterDefaultExpression instanceof Expr &&
+                    $parameterDefaultExpression instanceof  Expr &&
```
DanielBadura, line 60: This would introduce the php-parser as dep since the use statement. I could switch the check to `!== null` but we are still then relying on php-parser internals so i decided for now to check explicitly on the type.

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
```diff
@@ -38,8 +40,23 @@ public function __invoke(
         $changes = Changes::empty();
 
         foreach (Dict\intersect_by_key($fromParametersWithDefaults, $toParametersWithDefaults) as $parameterIndex => $parameter) {
-            $defaultValueFrom = $parameter->getDefaultValue();
-            $defaultValueTo   = $toParametersWithDefaults[$parameterIndex]->getDefaultValue();
+            try {
+                $defaultValueFrom = $parameter->getDefaultValue();
+                $defaultValueTo   = $toParametersWithDefaults[$parameterIndex]->getDefaultValue();
+            } catch (Throwable $throwable) {
```
Ocramius: We need a more precise exception type here

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
Ocramius: We would probably need php-parser for rendering the AST anyway, no? 🤔

Not happy about the added dependency, but it is what it is.

I will be very conservative on expression support though: don't want this to drift out of control 😁

test/unit/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChangedTest.php
```diff
@@ -71,7 +74,12 @@ function positionOfOptionalParameterChanged($a = 2, $b, $c = 1) {}
    class C {
        static function changed1($a = 1) {}
        function changed2($a = 1) {}
+       function notChangedNewInitializer($a = new stdClass()) {}
+       function notChangedNewEnum($a = D::A) {}
```
Ocramius, line 78: We need to test changes too

Ocramius (changes requested):

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
DanielBadura: Thought so, but the expetion is marked internal upstream so i hesitated to catch it explicitly.

Ocramius: we can improve it there, if needed

test/unit/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChangedTest.php
DanielBadura: Added two testcase which both dont report changes as expected

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
DanielBadura: I did narrow it down to the `UnableToCompileNode` exception.

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
DanielBadura: Yeah i can understand. I thought about only checking on `Expr_New` and `Expr_ClassConstFetch` as the type so we dont catch to much. WDYT?

DanielBadura:
Psalm seems for now not fixable since it complaining about undefined vars after the foreach but there is no possible case imho.

PHPcs will be fixed in the next commit round and infection as soon as we are finished polishing the behaviour.

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
Ocramius: It's fine to catch anything, IMO.

I would say that the current code is a bit too optimistic.

What about rendering the expression back to text, and then comparing there? That would be a bit safer, IMO.

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
```diff
@@ -38,8 +41,23 @@ public function __invoke(
...
+            } catch (UnableToCompileNode $unableToCompileNode) {
+                $parameterDefaultExpression   = $parameter->getDefaultValueExpression();
+                $toParameterDefaultExpression = $toParametersWithDefaults[$parameterIndex]->getDefaultValueExpression();
+
+                if (
+                    $toParameterDefaultExpression instanceof Expr &&
+                    $parameterDefaultExpression instanceof  Expr &&
+                    $toParameterDefaultExpression->getType() === $parameterDefaultExpression->getType()
+                ) {
+                    continue;
```
Ocramius, line 63: Let's rewrite this into a `private` function perhaps? Gets rid of the psalm issue too, I suppose.

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
DanielBadura: Finally having time to getting back to this:

I would also say it is too optimistic, that why i mentioned only catching the 2 types. But what do you mean by rendering it back to text? `getType` seemed to me the only possibilty to get a string like comparsion from `Expr`.

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
DanielBadura: Tried it right now, but this will not remove the the psalm issues. But i will still keep it in the private method.

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
Ocramius: > But what do you mean by rendering it back to text?

If the expressions change, we need to render the `Expr` as a `string` for the end-user to understand what has changed.

This should be done with the PHP-Parser PrettyPrinter, I suppose.

DanielBadura: Added now the prettyprinter and checking on the string retrieved from there. If this is fine from behaviour pov i would try satisfy the SA tools then e.g. getting psalm happy.

DanielBadura:
@Ocramius could you have a look again on this PR? Psalm errors were not fixable for me without defining the variables with `null`. Also moving the code chunk into a private method made the readabillity harder due to the mix of `continue` and `throw` in there.

Ocramius (approved):
Most excellent from my Pov, thanks @DanielBadura!

Ocramius:
Ah, one last thing: we'll need to rebase due to the `composer.json` and `composer.lock` changes. Will merge after that :+1:

DanielBadura:
@Ocramius PR is now rebased 👍

Ocramius (approved):
Thanks @DanielBadura!

PR #726, opened 2022-12-22
Update russian spelling

weirdan:
There are two similar but distinct adjectives in Russian, `русский` и `российский`. The former is used to refer to ethnicity/language/etc, and the latter is used to refer to the state/citizenship/etc (because historically Russia has been a multi-ethnic empire). 

Using `русским` was semantically wrong (reads as `To the ethnic Russian citizens (of an unknown country)`) and unnecessarily narrow at the same time.


Ocramius:
Note: this is also on all `laminas/` repositories: should it be adjusted there too? 🤔

weirdan:
> should it be adjusted there too?

Perhaps it should.

Ocramius:
Thanks @weirdan

PR #737, opened 2023-02-15
Support baseline configuration

lcobucci:
This allows users to set a list of regexes to filter known BC-breaks from detected changes.


README.md
```diff
@@ -121,4 +121,19 @@ vendor/bin/roave-backward-compatibility-check --help
 
 ## Configuration
 
-There are currently no configuration options available.
+The file `.roave-backward-compatibility-check.json` is read from the current working directory (when it exists) and sets configuration for the command.
+
+It's expected to be a JSON encoded file that, optionally, contains the following properties:
```
lcobucci: We can implement another format, I just think JSON is the simplest or most predictable...

README.md
```diff
@@ -121,4 +121,19 @@ vendor/bin/roave-backward-compatibility-check --help
 
 ## Configuration
 
-There are currently no configuration options available.
+The file `.roave-backward-compatibility-check.json` is read from the current working directory (when it exists) and sets configuration for the command.
```
lcobucci: We can also change the configuration file name. My goal was to not think too much about it while leaving room for adding new properties.

lcobucci:
@Nyholm although this uses a different name/format, it could be used to simplify https://github.com/Nyholm/roave-bc-check-docker :+1:

Ocramius:
I'll need to ponder on this: my preference would be to have either an XSD or a JSON schema, at least, and use those as documentation too 🤔

src/Command/AssertBackwardsCompatible.php
```diff
@@ -34,6 +35,8 @@
 
 final class AssertBackwardsCompatible extends Command
```
Ocramius: Would be a good idea to support `--generate-baseline` 🤔

lcobucci:
> I'll need to ponder on this: my preference would be to have either an XSD or a JSON schema, at least, and use those as documentation too 🤔 

I totally understand. I didn't want to introduce new dependencies to the project (I did implicitly add `ext-json`, though).

Up to you, mate!

src/Command/AssertBackwardsCompatible.php
lcobucci: I'd say it's a separate feature that may or not be added.

Since the introduction of BC-breaks isn't something done that often/lightly, I'd start with manual generation.

src/Baseline.php
```diff
@@ -0,0 +1,38 @@
...
+
+    public static function fromList(string ...$ignoredChanges): self
+    {
+        return new self(array_values($ignoredChanges));
+    }
+
+    public function ignores(Change $change): bool
+    {
+        foreach ($this->ignoredChanges as $ignoredChangeRegex) {
+            if (preg_match($ignoredChangeRegex, $change->__toString()) === 1) {
```
lcobucci: Not that it would make tons of difference but we just need to run `__toString()` once.

src/Baseline.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public static function fromList(string ...$ignoredChanges): self
+    {
+        return new self(array_values($ignoredChanges));
+    }
+
+    public function ignores(Change $change): bool
+    {
+        foreach ($this->ignoredChanges as $ignoredChangeRegex) {
```
lcobucci, line 33: We could compile (roughly) a single regex with all the items and reduce the regex match calls per detected changes.

It comes with a certain complexity, though.

src/Baseline.php
lcobucci: Updated :+1:

src/Baseline.php
asgrim: I can imagine if there are a huge amount of changes found and/or the regex list is huge, this could become a performance issue, but we can probably leave it as-is until it becomes a problem; it's understandable at least for now.

README.md
asgrim: I wonder if XML might be a nicer option, on account of not having to escape the regex?

src/Command/AssertBackwardsCompatible.php
```diff
@@ -213,4 +220,24 @@ private function determineFromRevisionFromRepository(
             $repository,
         );
     }
+
+    private function determineConfiguration(
```
asgrim: I'd advocate for separating this into a separate invokable method in a class; I can imagine this will grow in complexity if things like different filename support, validating config schema etc. get added, so isolating that initially _may_ be a nice idea?

src/Command/Configuration.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BackwardCompatibility\Command;
```
asgrim: This isn't a `Command` as such, it's a value object that holds configuration, maybe this should go in the top level for now?

```suggestion
namespace Roave\BackwardCompatibility;
```

src/Command/Configuration.php
```diff
@@ -0,0 +1,43 @@
...
+
+use Psl\Json;
+use Psl\Type;
+use Roave\BackwardCompatibility\Baseline;
+use RuntimeException;
+
+final class Configuration
```
asgrim: Do we need to make this `@psalm-immutable`, or does Psalm infer that these days? I genuinely don't know these days :grin:

asgrim (commented):
@lcobucci I'm loving this change, and it indeed formalises things that @Nyholm had done before (which I've used for a project too!) - really nice feature, looking forward to getting this shaped up :heart:

src/Baseline.php
lcobucci: I'd suggest leaving it as is for this iteration, honestly.

Compiling efficient regexes out of a variable list would also certainly impact performance and there's no way to guarantee we will have **one regex only**.

src/Command/AssertBackwardsCompatible.php
lcobucci: Sure thing, having it extracted now would certainly ease further extensions.

src/Command/Configuration.php
lcobucci: I can change it for sure.

My rationale was mostly focused on grouping together things that change together.
Even though the `Configuration` isn't a command it's very connected to commands and putting it closer makes things more explicit (to me at least :sweat_smile:).

It's the kind of class that I'd 100% mark as `private`, had we had that capability in PHP :smile:

src/Command/Configuration.php
lcobucci: I'm on the same boat but it doesn't hurt to add it :rofl:

README.md
lcobucci: I've implemented this quickly only to put the idea out there, JSON was the quickest thing I could use.

I'll change it to XML and add an XSD schema, my only question is: are you fine with the extra dependencies to be added to composer.json (ext-dom / ext-simplexml / ext-xmlreader / whateva)?

Ocramius: I'm fine with either, as long as there's a schema

src/Command/Configuration.php
lcobucci: I've decided to keep it as `Roave\BackwardCompatibility\Configuration\Configuration` since I still think it's a self-contained concept that isn't that important to be on the root level.

Happy to hear other arguments, though.

src/Command/AssertBackwardsCompatible.php
lcobucci: Alright, I had a few minutes today to play more with this and I've refactored things a bit on c251f5a to address this point (not yet implementing schema and all).

I've decided to keep the configuration determination logic in `src/Command` as, even though it isn't a command, it has dependencies to command stuff (`OutputInterface`).

Do you think it's sufficient for now? If so, I'll just replace the parsing/validation logic and we can call this a feature :smile:

test/unit/Command/ConfigurationTest.php
```diff
@@ -6,77 +6,16 @@
 
 use PHPUnit\Framework\TestCase;
```
lcobucci: I didn't move this file :man_facepalming: 
Oh well, something for the future me.

README.md
lcobucci: https://github.com/Roave/BackwardCompatibilityCheck/pull/737/commits/a7619cdd4144eb6c6811ce656aa0390f0404068a provides the XML+XSD implementation :+1:

Ocramius:
@lcobucci sorry, this is on my plate, didn't get to it though :(

lcobucci:
> @lcobucci sorry, this is on my plate, didn't get to it though :( 

Wow, the GH ui went crazy there... Just wanted to ask a review from you two (not to spam both).

No worries, we do what we can, whenever we can!

lcobucci:
That segfault seems to be unrelated... let me know if you need help with understanding why it's happening

README.md
asgrim: Docs need checking to change json to xml now

resources/schema.xsd
```diff
@@ -0,0 +1,24 @@
...
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+    <xs:annotation>
+        <xs:appinfo source="https://github.com/Roave/BackwardCompatibilityCheck"/>
+
+        <xs:documentation source="https://github.com/Roave/BackwardCompatibilityCheck">
+            This schema file defines the structure for the XML configuration file of roave/backward-compatibility-check.
+        </xs:documentation>
+    </xs:annotation>
+
+    <xs:element name="roave-bc-check" type="bcCheckType" />
```
asgrim: Can we be consistent with naming case (`<roave-bc-check>` vs `ignoredRegex`) please?

asgrim (commented):
No idea what the segfault is for, but it's happening consistently :grimacing:

README.md
lcobucci: Sure thing... I forgot it 🤦

resources/schema.xsd
lcobucci: Oh yeah, will do 👍

README.md
```diff
@@ -121,4 +121,20 @@ vendor/bin/roave-backward-compatibility-check --help
...
+
+It's expected to be an XML file that follows our [schema](resources/schema.xsd):
+
+**Example:**
+
+```xml
+<?xml version="1.0" encoding="UTF-8" ?>
+<roave-bc-check
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="resources/schema.xsd">
```
asgrim: I imagine this would be `vendor/roave/backward-compatibility-check/resources/schema.xsd` for a consumer, I'd think?

asgrim (approved):
Nice one! :heart: love this change

README.md
Ocramius: Yup

lcobucci: It depends™️ 😬 

If this lib has been installed using composer, yes.
If this lib has been installed using phar, not really... maybe the full URL would be better.

I was conflicted when I wrote that one...

resources/schema.xsd
```diff
@@ -0,0 +1,24 @@
...
+
+    <xs:complexType name="bcCheckType">
+        <xs:sequence>
+            <xs:element name="baseline" type="baselineType" minOccurs="0" />
```
Ocramius: Maximum once?

src/Command/DetermineConfigurationFromFilesystem.php
```diff
@@ -0,0 +1,36 @@
...
+use Psl\Type;
+use Roave\BackwardCompatibility\Configuration\Configuration;
+use Roave\BackwardCompatibility\Configuration\ParseConfigurationFile;
+use Roave\BackwardCompatibility\Configuration\ParseXmlConfigurationFile;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class DetermineConfigurationFromFilesystem
+{
+    public function __construct(
+        private readonly ParseConfigurationFile $parser = new ParseXmlConfigurationFile(),
```
Ocramius, line 18: Overall OK to do this, but it's gonna be fun to see all tooling break because of our first time using objects as default values 😁

src/Command/DetermineConfigurationFromFilesystem.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Roave\BackwardCompatibility\Command;
+
+use Psl\Str;
+use Psl\Type;
+use Roave\BackwardCompatibility\Configuration\Configuration;
+use Roave\BackwardCompatibility\Configuration\ParseConfigurationFile;
+use Roave\BackwardCompatibility\Configuration\ParseXmlConfigurationFile;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class DetermineConfigurationFromFilesystem
```
Ocramius, line 15: Maybe `@internal` here

src/Configuration/Configuration.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BackwardCompatibility\Configuration;
+
+use Roave\BackwardCompatibility\Baseline;
+
+/** @psalm-immutable */
```
Ocramius, line 9: Possibly `@internal` for now? 🤔

src/Configuration/InvalidConfigurationStructure.php
```diff
@@ -0,0 +1,32 @@
...
+
+use LibXMLError;
+use RuntimeException;
+
+use function sprintf;
+use function trim;
+
+use const PHP_EOL;
+
+final class InvalidConfigurationStructure extends RuntimeException
```
Ocramius, line 16: `@internal`

src/Configuration/ParseXmlConfigurationFile.php
```diff
@@ -0,0 +1,71 @@
...
+use DOMDocument;
+use Psl\File;
+use Roave\BackwardCompatibility\Baseline;
+use SimpleXMLElement;
+
+use function array_values;
+use function libxml_get_errors;
+use function libxml_use_internal_errors;
+
+final class ParseXmlConfigurationFile implements ParseConfigurationFile
```
Ocramius, line 17: `@internal`? 🤔

Ocramius (approved):
Looking good!

My only last concern is reducing public API as a start, perhaps, but I also don't mind bumping major version, if the API needs to change due to different directions taken in future 💪

README.md
Ocramius: The documentation example should likely assume `vendor/` path: `.phar` is really someone else's problem 😁

resources/schema.xsd
lcobucci: Yes, that's the default value for `xs:element` (this specific version, sure).

We can make it explicit

src/Command/DetermineConfigurationFromFilesystem.php
lcobucci: Why would it break? Unsupported?

resources/schema.xsd
Ocramius: Default works for me :+1:

src/Command/DetermineConfigurationFromFilesystem.php
Ocramius: Possibly in tooling, even in here. @DanielBadura has sent patches here to fix compatibility around that before, but it's really recent stuff :)

lcobucci:
All done :+1:

src/Configuration/Configuration.php
lcobucci: @Ocramius psalm isn't that happy due to the different top-level namespace for tests:

```sh
ERROR: InternalProperty - test/unit/Configuration/ConfigurationTest.php:18:47 - Roave\BackwardCompatibility\Configuration\Configuration::$baseline is internal to Roave but called from RoaveTest\BackwardCompatibility\Configuration\ConfigurationTest (see https://psalm.dev/176)
        self::assertEquals(Baseline::empty(), $config->baseline);


ERROR: InternalProperty - test/unit/Configuration/ConfigurationTest.php:19:26 - Roave\BackwardCompatibility\Configuration\Configuration::$filename is internal to Roave but called from RoaveTest\BackwardCompatibility\Configuration\ConfigurationTest (see https://psalm.dev/176)
        self::assertNull($config->filename);
```

Should we keep `@internal` and baseline these or should we remove `@internal`?

lcobucci: I've removed it for now.

Ocramius (approved):
Pristine job as usual, thanks @lcobucci! :gem:

Ocramius:
Hmm, the segfault part still requires a PHPUnit upgrade: checking that first then.

composer.json
```diff
@@ -26,9 +29,9 @@
     "require-dev": {
         "doctrine/coding-standard": "^11.0.0",
         "php-standard-library/psalm-plugin": "^2.2.1",
-        "phpunit/phpunit": "^9.5.27",
+        "phpunit/phpunit": "^10.0.14",
```
asgrim: Depends on #741 to be merged

test/e2e/Command/AssertBackwardsCompatibleTest.php
```diff
@@ -262,6 +273,21 @@ public function testWillPickLatestTaggedVersionOnNoGivenFrom(): void
...
+
+        $output = Shell\execute(__DIR__ . '/../../../bin/roave-backward-compatibility-check', [
+            '--from=' . $this->versions[0],
+            '--to=' . $this->versions[1],
+        ], $this->sourcesRepository, [], Shell\ErrorOutputBehavior::Append);
+
+        self::assertStringContainsString(
+            '.roave-backward-compatibility-check.xml" as configuration file',
+            $output,
+        );
```
azjezz, line 288: you can use `Shell\ErrorOutputBehavior::Packed` and `[$stdout, $stderr] = Shell\unpack($data)` to ensure that the string is in `stderr` not in `stdout`.

as of now, if the output goes to stdout instead of stderr, this test will still pass.

lcobucci:
@Ocramius alright, this should finally be good to go too.

I've updated the XSD to enforce a `#whateva#` pattern, which should be sufficient for pretty much all use cases.

Ocramius:
:ship:!

PR #762, opened 2024-03-12
Spring gardening: Symfony 7 support

michael-rubel:
# About

- Updates all the dependencies alongside Symfony Console v7.
- Updates tests, configurations, and fixes CI issues.

Symfony 7 is required for seamless usage with Laravel 11.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.idea
 /vendor
 /infection-log.txt
 /box.json
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 /vendor
 /infection-log.txt
 /box.json
--- a/composer.lock
+++ b/composer.lock
@@ -4,20 +4,20 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "6b34a9a8436c74e90966b5cf05ac02a9",
+    "content-hash": "5033c3b367a494da1a1564eb6b95172a",
     "packages": [
         {
             "name": "azjezz/psl",
-            "version": "2.3.1",
+            "version": "2.9.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/azjezz/psl.git",
-                "reference": "3ce5e7a24a0a5603ae482a0ec6d4aaf10311c45c"
+                "reference": "a8685b297644f3898c986e89b15309d31b1f47bf"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/azjezz/psl/zipball/3ce5e7a24a0a5603ae482a0ec6d4aaf10311c45c",
-                "reference": "3ce5e7a24a0a5603ae482a0ec6d4aaf10311c45c",
+                "url": "https://api.github.com/repos/azjezz/psl/zipball/a8685b297644f3898c986e89b15309d31b1f47bf",
+                "reference": "a8685b297644f3898c986e89b15309d31b1f47bf",
                 "shasum": ""
             },
             "require": {
@@ -26,18 +26,18 @@
                 "ext-json": "*",
                 "ext-mbstring": "*",
                 "ext-sodium": "*",
-                "php": "~8.1.0 || ~8.2.0",
-                "revolt/event-loop": "^1.0.0"
+                "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
+                "revolt/event-loop": "^1.0.1"
             },
             "require-dev": {
-                "friendsofphp/php-cs-fixer": "^3.4.0",
-                "php-coveralls/php-coveralls": "^2.5.2",
-                "php-standard-library/psalm-plugin": "^2.1.0",
-                "phpbench/phpbench": "^1.2.3",
-                "phpunit/phpunit": "^9.5.16",
-                "roave/infection-static-analysis-plugin": "^1.26.0",
-                "squizlabs/php_codesniffer": "^3.6.2",
-                "vimeo/psalm": "^5.0.0"
+                "friendsofphp/php-cs-fixer": "^3.22.0",
+                "php-coveralls/php-coveralls": "^2.6.0",
+                "php-standard-library/psalm-plugin": "^2.2.1",
+                "phpbench/phpbench": "^1.2.14",
+                "phpunit/phpunit": "^9.6.10",
+                "roave/infection-static-analysis-plugin": "^1.32.0",
+                "squizlabs/php_codesniffer": "^3.7.2",
+                "vimeo/psalm": "^5.13.1"
             },
             "suggest": {
                 "php-standard-library/psalm-plugin": "Psalm integration"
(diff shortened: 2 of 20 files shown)


michael-rubel:
Hey @Ocramius 

Can we do anything to allow symfony/console ^7.0 without breaking the CI, or would you rather address the newly reported issues?

Ocramius:
@michael-rubel I suggest trying a very minimal `composer update symfony/console` to reduce `composer.lock` changes.

Also, let's please **bump** the range: don't widen it.

composer.lock
```diff
@@ -1611,24 +1611,24 @@
         },
         {
             "name": "symfony/console",
-            "version": "v6.2.3",
+            "version": "v6.4.4",
```
Ocramius: Something is holding the version back

michael-rubel:
Yeah, I tried that minimal update at first but psalm and some other packages want to upgrade before it installs v7.

Ocramius:
Would suggest helping with #732 first then :)

michael-rubel:
It looks like the bot already tried the update in https://github.com/Roave/BackwardCompatibilityCheck/pull/753

The https://github.com/Roave/BackwardCompatibilityCheck/pull/732 is a bit stale. It probably makes sense to bump the lock to the latest versions and then address the issues, because for example psalm is required at least from 5.16 as it's the version when they added support for Symfony 7 (ref: https://github.com/vimeo/psalm/releases/tag/5.16.0).

Ocramius:
That was already attempted in #730: needs work :)

michael-rubel:
Hell, I'll see if I have spare time for this.

Ocramius:
The project needs some spring gardening work, basically :D

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
Ocramius: This should be in your global gitignore, not here  ;-)

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
```diff
@@ -62,8 +62,6 @@ public function __invoke(
                 ) {
                     continue;
                 }
-
-                throw $unableToCompileNode;
```
Ocramius: I'd expect this re-throw to stay? :thinking:

michael-rubel:
@Ocramius Should we drop PHP 8.1 for Symfony 7 support?

michael-rubel:
@Ocramius It seems that everything passes on PHP 8.2+

src/Change.php
```diff
@@ -77,6 +77,7 @@ public function withFilePositionsIfNotAlreadySet(
         int|null $line,
         int|null $column,
     ): self {
+        /** @infection-ignore-all */
```
Ocramius: Should not ignore here: need to dig down to why infection adds mutations here :thinking:

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
```diff
@@ -63,6 +63,7 @@ public function __invoke(
                     continue;
                 }
 
+                /** @infection-ignore-all */
```
Ocramius: Should not ignore here: need to dig down to why infection adds mutations here :thinking:

src/SourceLocator/LocatedSourceWithStrippedSourcesDirectory.php
```diff
@@ -10,7 +10,11 @@
 use function strlen;
 use function substr_replace;
 
-/** @internal */
+/**
+ * @internal
+ *
+ * @psalm-immutable
```
Ocramius, line 16: This object depends on `LocatedSource` being immutable: is it immutable, or are we lying? :thinking:

Ocramius (commented):
Looking good! I haven't dug in depth, since I'm multitasking like crazy, but I think we should hunt down the last few suppressions first.

I know it's annoying, but it takes so many days to recover from going below threshold @_@

src/Change.php
michael-rubel: This one is because it can remove the `clone` keyword. I wasn't sure what was the intention to do that.

michael-rubel: This one is because it can remove the `clone` keyword. I wasn't sure what was the intention to use it.

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
michael-rubel: The exception wasn't expected in the tests, so it tries to remove the throw.

src/SourceLocator/LocatedSourceWithStrippedSourcesDirectory.php
michael-rubel: The class only contains getters and boolean checks. Or am I missing something?

src/Change.php
michael-rubel: This one is because it can remove the clone keyword.

Ocramius: @michael-rubel yeah, this kind of signaling (infection removing stuff) indicates that we have a test to add, BTW :)

src/DetectChanges/BCBreak/FunctionBased/ParameterDefaultValueChanged.php
Ocramius: We likely need to see if there's a test scenario to be written here

src/SourceLocator/LocatedSourceWithStrippedSourcesDirectory.php
Ocramius: Is the inner located source also immutable? That's what decides if this one is immutable too :D

michael-rubel: The base class looks similar. What's wrong here?

src/Change.php
michael-rubel: Yeah, I was unsure if it makes sense to cover.

src/SourceLocator/LocatedSourceWithStrippedSourcesDirectory.php
michael-rubel: The base class looks similar. What's wrong there?

michael-rubel:
@Ocramius Ping 👋

Ocramius:
@michael-rubel I need to let this sit for a bit: currently stuck with pre-easter-vacation-deadlines-everywhere. Do ping again if stuck after Easter!

michael-rubel:
@Ocramius how's it going? 😁

Ocramius:
Only one failure, and it's not really part of the public API of this package anyway: nice! 

```
Error: The number of required arguments for Symfony\Component\Console\Command\Command#setApplication() increased from 0 to 1
```

Let's :ship: this

src/SourceLocator/LocatedSourceWithStrippedSourcesDirectory.php
Ocramius: Just clarifying here: I thought I saw a suppression, but there wasn't one :-) 

Possibly due to older `roave/better-reflection` dependency before your patch :+1:

PR #768, opened 2024-05-19
Detect changes in ENUMs as BC breaks

### LINKED ISSUE #767: New feature: Detect EnumCaseAdded as a BC break
As far as I can tell this isn't something currently detected, but I think adding a new enum case should be considered a breaking change, as it will break any consuming code that was previously relying on being able to exhaustively handle all cases of an enum.

For instance consuming code will often use match with an enum, which will throw [UnhandledMatchError ](https://www.php.net/manual/en/class.unhandledmatcherror.php) if the enum cases is a new one.

Would people be interested in a PR to implement this check?

bdsl:
See https://github.com/Roave/BackwardCompatibilityCheck/issues/767

I went slightly further than discussed in the issue, reporting a BC break for cases removed, as well as for cases added. In both cases cases marked `@internal` do not trigger the BC break.

I think there's more to do for a complete handling of enums, including detecting if an entire Enum is removed, or becomes a trait, class or interface. Actually I see the entire Enum removed is already handled by the catch of `IdentifierNotFound` 

https://github.com/Roave/BackwardCompatibilityCheck/blob/4c05d1a5a164ae21d5d0f6f994b116b7f71323bd/src/CompareClasses.php#L76

and I could easily extend this PR to handle became trait/interface/class by simply treating one of those as a thing that has no cases - so an enum becoming one of those is not an issue itself, its only an issue because it means cases go away. And an enum with no cases becoming a different sort of classlike is not a BC break. Let me know if you'd like me to do that.


bdsl:
If you approve the workflow I'll try to fix any issues it detects.

I've run unit tests, psalm and phpcs on my local. It might be nice to have some end to end testing as well though. For this sort of project I wish git would work recursively, letting us have a git repo of a sample project inside this repo's test assets directory.

Ocramius:
@bdsl done

bdsl:
I also noticed there seems to be quite a bit of code duplication in the existing code between the handling of Classes, Traits and Interface. I think there could be a nice way to remove that duplication, but I didn't want to get into restructuring the existing code on my first PR. E.g. the `ClassBased`,  `TraitBased`, `InterfaceBased` and new `EnumBased` interfaces are all the same except with differently named parameters, and the three `MultipleChecksOnA...` classes are also extremely similar. I didn't feel like writing `MultipleChecksOnAnEnum` which is partly why I implemented all the checks within one class `EnumBased\ClassChanged`

bdsl:
ah didn't realise you need to approve the workflow separately for every commit. I've turned on github actions in my fork so we can see the results there: https://github.com/bdsl/BackwardCompatibilityCheck/actions

bdsl:
The Infection check is currently failing. But on my local using ` vendor/bin/roave-infection-static-analysis-plugin  --threads=8 --logger-html=infection.html` I'm now getting 0 undetected mutants, so I'm not quite sure why. I'll have to come back to this later to see how to reproduce the way that github actions is running infection.

bdsl:
@Ocramius can you see where I'm going wrong with the mutation tests? On my laptop (php 8.3 with xdebug in WSL) I'm getting 0 returned from the infection command as shown below. I don't know if it could be because I'm running xdebug rather than phpdbg.

```
$ git show --stat; git status; ./vendor/bin/roave-infection-static-analysis-plugin; echo $?
commit b5cfc9b2e141b0cd772db8e5f7ca95d9cf68a898 (HEAD -> 767-enum-cases, bdsl/767-enum-cases)
Author: Barney Laurance <barney@redmagic.org.uk>
Date:   Sun May 19 17:34:06 2024 +0100

    Roave/BackwardCompatibilityCheck#767: Remove unecassary array_values calls
    
    Infection detected that these were not needed

 src/DetectChanges/BCBreak/EnumBased/CasesChanged.php | 9 +++++----
 1 file changed, 5 insertions(+), 4 deletions(-)
On branch 767-enum-cases
nothing to commit, working tree clean

    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

#StandWithUkraine

Infection - PHP Mutation Testing Framework version 0.27.10


Running initial test suite...

PHPUnit version: 9.6.18

  755 [============================] 7 secs

Generate mutants...

Processing source code files: 115/115
.: killed, M: escaped, U: uncovered, E: fatal error, X: syntax error, T: timed out, S: skipped, I: ignored

..................................................   ( 50 / 639)
..................................................   (100 / 639)
..................................................   (150 / 639)
..................................................   (200 / 639)
..................................................   (250 / 639)
..................................................   (300 / 639)
..................................................   (350 / 639)
..................................................   (400 / 639)
..................................................   (450 / 639)
..................................................   (500 / 639)
..................................................   (550 / 639)
....E.............................................   (600 / 639)
.......................................              (639 / 639)

639 mutations were generated:
     638 mutants were killed
       0 mutants were configured to be ignored
       0 mutants were not covered by tests
       0 covered mutants were not detected
       1 errors were encountered
       0 syntax errors were encountered
       0 time outs were encountered
       0 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 100%
         Mutation Code Coverage: 100%
         Covered Code MSI: 100%

Note: to see escaped mutants run Infection with "--show-mutations" or configure file loggers.

Please note that some mutants will inevitably be harmless (i.e. false positives).
Escaped mutants:
================

Timed Out mutants:
==================

Skipped mutants:
================

Not Covered mutants:
====================
[warning] Dashboard report has not been sent: The current process is not executed in a CI build
 ! [NOTE] The MSI is 9.9% percentage points over the required MSI. Consider increasing the required MSI percentage the  
 !        next time you run Infection.


Time: 55s. Memory: 0.11GB. Threads: 1
0
```

Ocramius:
> Notice:  You are running Infection with phpdbg enabled.

Perhaps it's using `phpdbg` for coverage, rather than XDebug

bdsl:
> > Notice:  You are running Infection with phpdbg enabled.
> 
> Perhaps it's using `phpdbg` for coverage, rather than XDebug

Yes possibly. I'll see if I can it working with phpdbg on my local.

bdsl:
Spent some more time looking at this but I didn't find a way to reproduce the github actiosn behaviour on my local.

I did notice that the head on the origin branch also is showing some uncovered mutants on the action output ( https://github.com/Roave/BackwardCompatibilityCheck/actions/runs/8652266041/job/23725019284 ) although not enough to reduce the MSI below 90% and so not enough count as a build failure. But on my local with 8.2 and xdebug I get 100% MSI. (there's some crash with phpdbg )

My changes are reducing the MSI as measured in github from 92% to 88%.

I'll try and take a look again the next few days if no-one else works out what's going on.

Ocramius:
@bdsl thanks for checking: could also be caused by underlying PHP updates, then (the rest is in `composer.lock`).

I'd suggest that we do a round of reviews, and then consider whether we should lower MT requirements for this to go through, as it's possibly well tested already :-)

bdsl:
> @bdsl thanks for checking: could also be caused by underlying PHP updates, then (the rest is in `composer.lock`).
> 
> I'd suggest that we do a round of reviews, and then consider whether we should lower MT requirements for this to go through, as it's possibly well tested already :-)

@Ocramius Sounds good. No rush.

I'm also wondering if it might be an issue that needs fixing in [laminas/laminas-continuous-integration-action](https://github.com/laminas/laminas-continuous-integration-action) about how it collects line coverage stats to feed into infection.

src/CompareClasses.php
```diff
@@ -20,6 +21,7 @@ public function __construct(
         private ClassBased $classBasedComparisons,
         private InterfaceBased $interfaceBasedComparisons,
         private TraitBased $traitBasedComparisons,
+        private EnumBased|null $enumBasedComparisons = null, // optional argument for now to avoid BC break
```
Ocramius: I'd suggest merging these into `ClassBased`: that removes all code changes from `Compareclasses`

src/DetectChanges/BCBreak/EnumBased/EnumBased.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BackwardCompatibility\DetectChanges\BCBreak\EnumBased;
+
+use Roave\BackwardCompatibility\Changes;
+use Roave\BetterReflection\Reflection\ReflectionClass;
+
+interface EnumBased
```
Ocramius: While this interface makes sense, I would make it extend from `ClassBased`, since ENUMs in PHP are just `final` classes

Ocramius (changes requested):
The changes are simple and minimal.

I would suggest removing the `EnumBased` additional argument for `CompareClasses`, and move the `EnumBased\CasesChanged()` directly into the `ClassBased` checks.

test/asset/api/old/EnumWithCasesBeingChanged.php
```diff
@@ -0,0 +1,24 @@
...
+    case March;
+    case April;
+    case May;
+    case June;
+    case July;
+    case August;
+    case September;
+    case October;
+    case November;
+    Case December;
```
Ocramius: What about CasEsensiTIVitY issues?

bdsl:
Thanks for the review @Ocramius - I'll try and find time to make changes tomorrow. Treating an enum as just a type of class does make things simpler, I was just treating them separately to follow the example of the other classlikes. Happy to change that.

I might add the check for CaseBecameInternal at the same time.

Do you have any objection to me also covering the case of an enum being changed to a different classlike by simply treating any non-enum classlike as simply a thing that has no cases. So the change is a BC break iff there is at least one non-internal case in the old enum.

test/asset/api/old/EnumWithCasesBeingChanged.php
bdsl: Case names are case sensitive, but I'm not sure what issue you see with that.  I'm using `\Roave\BetterReflection\Reflection\ReflectionEnum::hasCase` which is use the case sensitive `array_key_exists` . I can't see any code that means a combination of cases with different names that are equal under case-insensitive comparison needs any different handling to any other combination of cases with different names.

src/DetectChanges/BCBreak/EnumBased/EnumBased.php
bdsl: If I remove the change to CompareClasses as you suggested and I'm planning to do then I'd just delete this interface and use ClassBased. It doesn't have any significant distinction.

The only difference between this and ClassBased is the different param names, which I think would mean SA tools will rightly say it's not allowed to extend ClassBased.

test/asset/api/old/EnumWithCasesBeingChanged.php
Ocramius: @bdsl I was thinking about renaming `December` to `december` (as an example) - does that result in a BC issue?

bdsl: Right. I'll test it. I expect it will be reported as two BC issues, removal of `December` and addition of `december`. That seems like correct behaviour.

Ocramius: It's just about adding that difference to this existing test, BTW :+1:

test/unit/CompareClassesTest.php
```diff
@@ -153,7 +153,6 @@ public function testAnonymousClassesAreFilteredOut(): void
         $this->classBasedComparatorWillNotBeCalled();
         $this->interfaceBasedComparatorWillNotBeCalled();
         $this->traitBasedComparatorWillNotBeCalled();
```
bdsl:
```suggestion
        $this->traitBasedComparatorWillNotBeCalled();
        
```

src/CompareClasses.php
bdsl: Done.

src/DetectChanges/BCBreak/EnumBased/EnumBased.php
bdsl: Interface removed.

test/asset/api/old/EnumWithCasesBeingChanged.php
bdsl: Added a change of case of case to the test.

bdsl:
This is ready for re-review.

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
```diff
@@ -0,0 +1,162 @@
...
+use Roave\BackwardCompatibility\Change;
+use Roave\BackwardCompatibility\Changes;
+use Roave\BetterReflection\Reflection\ReflectionClass;
+use Roave\BetterReflection\Reflection\ReflectionEnum;
+use Roave\BetterReflection\Reflection\ReflectionEnumCase;
+
+use function array_filter;
+use function array_map;
+
+class EnumCasesChanged implements ClassBased
```
Ocramius: `final`

test/unit/DetectChanges/BCBreak/ClassBased/EnumCasesChangedTest.php
```diff
@@ -0,0 +1,97 @@
...
+use Roave\BetterReflection\BetterReflection;
+use Roave\BetterReflection\Reflection\ReflectionClass;
+use Roave\BetterReflection\Reflector\DefaultReflector;
+use Roave\BetterReflection\SourceLocator\Type\SingleFileSourceLocator;
+use stdClass;
+
+use function array_map;
+use function iterator_to_array;
+
+final class EnumCasesChangedTest extends TestCase
```
Ocramius, line 20: `@covers` to be added here

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
```diff
@@ -0,0 +1,162 @@
...
+        $fromKind     = $this->kindOf($fromClass);
+        $toKind       = $this->kindOf($toClass);
+
+        if (! $fromClass instanceof ReflectionEnum && ! $toClass instanceof ReflectionEnum) {
+            return Changes::empty();
+        }
+
+        if (! $fromClass instanceof ReflectionEnum && $toClass instanceof ReflectionEnum) {
+            return Changes::fromList(Change::changed($fromKind . ' ' . $fromEnumName . ' became enum'));
+        }
```
Ocramius: The second `&&` component can be removed, in theory?

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
```diff
@@ -0,0 +1,162 @@
...
+        $fromEnumName = $fromClass->getName();
+        $fromKind     = $this->kindOf($fromClass);
+        $toKind       = $this->kindOf($toClass);
+
+        if (! $fromClass instanceof ReflectionEnum && ! $toClass instanceof ReflectionEnum) {
+            return Changes::empty();
+        }
+
+        if (! $fromClass instanceof ReflectionEnum && $toClass instanceof ReflectionEnum) {
+            return Changes::fromList(Change::changed($fromKind . ' ' . $fromEnumName . ' became enum'));
```
Ocramius, line 30: Is a `class` becoming an `enum` really a BC break, if all of its API is preserved? :thinking: 

Perhaps it is because of `new` no longer working? I'm mostly fine with this change, but just noticed, and am challenging my reasoning :D

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
```diff
@@ -0,0 +1,162 @@
...
+            ...$caseRemovedChanges,
+            ...$caseAddedChanges,
+            ...$caseBecameInternalChanges,
+            ...$caseBecameNotInternalChanges,
+        );
+    }
+
+    /**
+     * Copied from DetectChanges\BCBreak\ClassBased\ExcludeInternalClass - for now I'm not sure
+     * if there's a good place to put a shared function, and following the 3 strike then refactor rule.
```
Ocramius: Comment can go away, for now: an `isInternal` helper is certainly useful, but not needed for this patch :)

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
```diff
@@ -0,0 +1,162 @@
...
+            },
+            $nowNotInternalCases,
+        );
+
+        return Changes::fromList(
+            ...$caseRemovedChanges,
+            ...$caseAddedChanges,
+            ...$caseBecameInternalChanges,
+            ...$caseBecameNotInternalChanges,
+        );
```
Ocramius, line 132: Overall OK for this patch.

If you happen to touch this repo (in a different feature) again, though, I endorse splitting such various cases in a more detailed way, since you covered "everything enum" in one go, which is a bit much :-)

Ocramius: Specifically, this allows configuration by downstream consumers:

`new EnumCasesChanged` vs `new EnumBecameNonEnum` vs `NonEnumBecameEnum` or such

Ocramius (approved):
`final` to be added, plus some minor adjustments, then this can :ship:  :)

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
bdsl:
```suggestion
final class EnumCasesChanged implements ClassBased
```

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
bdsl: Yes, I think class and enum are different enough that it's very unlikely for there to be a useful case way to make that change that isn't a bc break:

- New doesn't work as you say. Not a bc break if the constructor was private or internal, but checking for those is more complexity.
- De-serialisation doesn't work. Again might not be a BC break if it didn't before but would be something else to check for.
- I think the original reason I assumed it would be a BC break is because it means all cases must be new (unless it's a enum with no cases but that's hard to think of a purpose for), and we said already a new case is a BC break. Maybe there's an exception could be made when the old version had zero cases, but again probably complexity we don't need.
- An enum can't be extended, so a BC break if the class was non-final.

Maybe others.

Ocramius: Excellent, thanks for confirming!

And indeed, didn't think of (de-)serialization! Good catch!

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
bdsl: > Overall OK for this patch.
> 
> If you happen to touch this repo (in a different feature) again, though, I endorse splitting such various cases in a more detailed way, since you covered "everything enum" in one go, which is a bit much :-)

Yes, I did think this might have grown a bit much. It's partly because I started with the EnumBased as an interface etc, and didn't want to copy the logic for that sort of thing from ClassBased.

So also curious what you'd think about a refactor to reduce some of the duplicate / similar code between ClassBased, InterfaceBased and TraitBased

Ocramius: > So also curious what you'd think about a refactor to reduce some of the duplicate / similar code between ClassBased, InterfaceBased and TraitBased

About this, I think "categorizing types of breakages" is more valuable than preventing duplication.

I did not work much on this repo in the last years, but my general idea is to improve the report with:

 * the code location of a **specific** breakage
 * a trace of a breakage: nested rules should be displayed, such as `non-final class -> method added` or `interface -> method added`
 * an explanation for every BC rule of why a specific rule is a breakage, so that users can be informed about reasoning, not just problems

That all leads to a lot of duplication that is still valuable, though :)

test/unit/DetectChanges/BCBreak/ClassBased/EnumCasesChangedTest.php
bdsl: Done in 5519e97b858f

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
bdsl: Good catch, done in 190968c0a5cbd7

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
bdsl: Done in 6703c313f328

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
Ocramius: BTW, if the Infection report was more precise, this is precisely the kind of stuff it catches :D

test/unit/DetectChanges/BCBreak/ClassBased/EnumCasesChangedTest.php
```diff
@@ -0,0 +1,98 @@
...
+use Roave\BetterReflection\BetterReflection;
+use Roave\BetterReflection\Reflection\ReflectionClass;
+use Roave\BetterReflection\Reflector\DefaultReflector;
+use Roave\BetterReflection\SourceLocator\Type\SingleFileSourceLocator;
+use stdClass;
+
+use function array_map;
+use function iterator_to_array;
+
+/** @covers EnumCasesChanged */
```
Ocramius: `@covers` always wants a FQCN, sorry @_@ 

```suggestion
/** @covers \Roave\BackwardCompatibility\DetectChanges\BCBreak\ClassBased\EnumCasesChanged */
```

bdsl: Thanks. I probably should have taken the hint from PHPStorm writing it as a FQCN and not offering to shorten it.

Ocramius:
All looking quite good!

I'll do a local check of specific mutants before merging :-)

Ocramius:
Took me a bit (because I just switched my OS / main computer), but this is **pristine** when running with XDebug!

```
692 mutations were generated:
     691 mutants were killed
       0 mutants were configured to be ignored
       0 mutants were not covered by tests
       0 covered mutants were not detected
       1 errors were encountered
       0 syntax errors were encountered
       0 time outs were encountered
       0 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 100%
         Mutation Code Coverage: 100%
         Covered Code MSI: 100%

Note: to see escaped mutants run Infection with "--show-mutations" or configure file loggers.

Please note that some mutants will inevitably be harmless (i.e. false positives).
Escaped mutants:
================

Timed Out mutants:
==================

Skipped mutants:
================

Not Covered mutants:
====================
[warning] Dashboard report has not been sent: The current process is not executed in a CI build
 ! [NOTE] The MSI is 9.9% percentage points over the required MSI. Consider increasing the required MSI percentage the  
 !        next time you run Infection.
```

I will lower the MT requirements to get this merged, meanwhile :-)

bdsl:
> Took me a bit (because I just switched my OS / main computer), but this is **pristine** when running with XDebug!

Yep, I ran infection on my local when I saw that it was failing on the build, so then killed all the mutants.

There are some mutations I can imagine easily that are not tested, like deleting the untested "return trait" here or equivalently replacing the if condition with `false`. But I think enough is tested. 

image

Ocramius:
Meanwhile, :ship: 

Thanks @bdsl!

src/DetectChanges/BCBreak/ClassBased/EnumCasesChanged.php
bdsl: > > So also curious what you'd think about a refactor to reduce some of the duplicate / similar code between ClassBased, InterfaceBased and TraitBased
> 
> About this, I think "categorizing types of breakages" is more valuable than preventing duplication.
> 
> I did not work much on this repo in the last years, but my general idea is to improve the report with:
> 
>     * the code location of a **specific** breakage
> 
>     * a trace of a breakage: nested rules should be displayed, such as `non-final class -> method added` or `interface -> method added`
> 
>     * an explanation for every BC rule of why a specific rule is a breakage, so that users can be informed about reasoning, not just problems
> 
> 
> That all leads to a lot of duplication that is still valuable, though :)

I'm not sure, I think you could still have all the categorisation and keep the exact same output without the code duplication. Maybe with an enum for the category!  I might try that refactor if/when I have another contribution to make.

Ocramius: I'm fine with any refactoring that improves things, but be aware that besides some duplication, this library has little to no known problems :D

PR #829, opened 2024-10-07
Add support for `JsonFormatter`

Nyholm:
This is useful when other programs wants to read the output data.

### DIFF
--- a/src/Formatter/JsonFormatter.php
+++ b/src/Formatter/JsonFormatter.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BackwardCompatibility\Formatter;
+
+use Psl\Str;
+use Roave\BackwardCompatibility\Changes;
+use Roave\BackwardCompatibility\Git\CheckedOutRepository;
+use Symfony\Component\Console\Output\OutputInterface;
+
+use function json_encode;
+
+/** @internal */
+final class JsonFormatter implements OutputFormatter
+{
+    public function __construct(
+        private readonly OutputInterface $output,
+        private CheckedOutRepository $basePath,
+    ) {
+    }
+
+    public function write(Changes $changes): void
+    {
+        $basePath = $this->basePath->__toString() . '/';
+        $result   = [];
+
+        foreach ($changes as $change) {
+            $result[] = [
+                'description' => $change->description,
+                'path' => $change->file === null ? null : Str\replace($change->file, $basePath, ''),
+                'line' => $change->line,
+                'column' => $change->column,
+            ];
+        }
+
+        $this->output->writeln(json_encode(['errors' => $result]));
+    }
+}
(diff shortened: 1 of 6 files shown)


test/unit/Formatter/JsonFormatterTest.php
```diff
@@ -0,0 +1,83 @@
...
+                ['description'=>'file-in-checked-out-dir','path'=>'subpath/file-in-checked-out-dir.php','line'=>10,'column'=>20],
+            ]
+        ];
+
+        $result = $output->fetch();
+        $data = json_decode(trim($result), true);
+        $this->assertIsArray($data);
+        $this->assertEquals($expected, $data);
+
+        self::assertEquals(
```
Ocramius: Possibly worth using PHPUnit's built-in Json assertions here: it gives much better error messages, if anything shifts for whatever reason

src/Formatter/JsonFormatter.php
```diff
@@ -0,0 +1,40 @@
...
+
+use Psl\Str;
+use Roave\BackwardCompatibility\Changes;
+use Roave\BackwardCompatibility\Formatter\OutputFormatter;
+use Roave\BackwardCompatibility\Git\CheckedOutRepository;
+use Symfony\Component\Console\Output\OutputInterface;
+use function Psl\Str;
+
+/** @internal */
+final class JsonFormatter implements OutputFormatter
```
Ocramius, line 15: We should ship an accompanying JSON Schema file with this, if we want consumers downstream to rely on it. It suffices for it to be in the docs, although verifying the schema in the test suite would be preferable.

Ocramius (changes requested):

Nyholm:
Thank you for the quick review and the suggestions. 

I've updated the PR accordingly.

test/unit/Formatter/JsonFormatterTest.php
```diff
@@ -76,7 +76,7 @@ public function testWrite(): void
         self::assertIsArray($data);
         self::assertEquals($expected, $data);
 
-        JsonAssert::assertJsonMatchesSchema($json, dirname(__DIR__, 3) . '/Resources/errors.schema.json');
+        JsonAssert::assertJsonMatchesSchema(json_decode($json), dirname(__DIR__, 3) . '/Resources/errors.schema.json');
```
Ocramius: I think `JSON_THROWS_ON_ERROR` is needed here, no? :thinking:

Ocramius: Ok, the library is a bit weird here, since it does not seem to accept primitive values at all :D 

`assert::isArray()` would suffice before this statement here, so it `array<...>` can be inferred first

Nyholm: > I think JSON_THROWS_ON_ERROR is needed here, no? 

Sure, 

> assert::isArray() would suffice before this statement here, so it array<...> can be inferred first

It is a `stdClass`, Using an array will not work :/, The library cannot tell the difference between an object and array.

asgrim (approved):
Nice! I like it, and well defined too with a schema :100:

Ocramius (approved):
:ship: then, thanks @Nyholm!

Nyholm:
Thank you for merging

Ocramius:
Is also released :)

PR #984, opened 2026-02-02
Fix schema location case sensitivity in README.md example

alexander-schranz:
PHPStorm on Mac seems else not detect the path correctly.


Ocramius (approved):
Thanks @alexander-schranz!

PR #986, opened 2026-02-09
bug: `protected` method in final class -> `private` is not a BC break

acoulton:
A method should only be considered "accessible" outside the class if:
 
* It is `public`
* It is `protected` and the class is NOT `final`

Previously, `protected` methods in `final` classes were incorrectly 
considered to be accessible. This meant that changing one to `protected`
would incorrectly be reported as a BC break.

This PR introduces two commits - one that introduces a new testcase to prove the issue, and a second with the fix. I will push separately so that CI shows the new test failing initially.

### DIFF
--- a/test/unit/DetectChanges/BCBreak/ClassBased/MethodRemovedTest.php
+++ b/test/unit/DetectChanges/BCBreak/ClassBased/MethodRemovedTest.php
@@ -13,6 +13,7 @@ use Roave\BetterReflection\BetterReflection;
 use Roave\BetterReflection\Reflection\ReflectionClass;
 use Roave\BetterReflection\Reflector\DefaultReflector;
 use Roave\BetterReflection\SourceLocator\Type\SingleFileSourceLocator;
+use Roave\BetterReflection\SourceLocator\Type\StringSourceLocator;
 
 use function array_map;
 use function iterator_to_array;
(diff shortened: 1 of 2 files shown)


acoulton:
I realised my "push separately" strategy doesn't work as the CI needs approval for new contributors. I figured it was therefore easier to push both commits now, but let me know if you'd like to see proof of the test failing.

test/unit/DetectChanges/BCBreak/ClassBased/MethodRemovedTest.php
```diff
@@ -60,6 +61,48 @@ public static function classesToBeTested(): array
...
+                    <<<'PHP'
+                    <?php
+                    final class FinalClassWithProtectedMethod {
+                        private function foo(string $bar): void {}
+                    }
+                    PHP,
+                    $locator,
+                )))->reflectClass('FinalClassWithProtectedMethod'),
+                [],
+            ],
```
acoulton, line 84: I saw some of the tests (including the existing case for this test) use external fixture files, while others use inline strings. I thought it was easier to show the issue here with inline strings but let me know if you'd like them extracted to fixtures instead.

asgrim (approved):

test/unit/DetectChanges/BCBreak/ClassBased/MethodRemovedTest.php
Ocramius: It's fine with inline fixtures, thanks! 👍

Ocramius (approved):
🚢 thanks @acoulton!

PR #987, opened 2026-02-13
fix: adding optional parameters to contructor

### LINKED ISSUE #978: False-positive when adding optional parameter to constructor
Example: https://github.com/yiisoft/log/pull/128

Before:

```php
public function __construct()
{
    // ...
}
```

After:

```php
public function __construct(array $levels = [])
{
    // ...
}
```

Result:

```
[BC] ADDED: Parameter levels was added to Method __construct() of class Yiisoft\Log\Target
```

mstegmeyer:
Adding optional parameters to a constructor is not a breaking change, since child classes do not need to match the signature of the parent. Only non-optional parameters (or parameters with default values before the last required parameter) are a breaking change in a constructor, as here the method call would need to be adjusted

### Small example

```php
class A {
    public function __construct(int $one, int $two = 666) {}
}

class B extends A {
    public function __construct(string $whatever) {
        parent::__construct(1);
    }
}

new B('foo');
new A(2);
```

Adding the parameter `$two` in class `A` here does not have an effect on neither constructing `A` nor `B`.

You can try it out here: https://3v4l.org/QYsC8

fixes https://github.com/Roave/BackwardCompatibilityCheck/issues/978

### DIFF
--- a/src/DetectChanges/BCBreak/MethodBased/MethodParameterAdded.php
+++ b/src/DetectChanges/BCBreak/MethodBased/MethodParameterAdded.php
@@ -33,12 +33,7 @@ final class MethodParameterAdded implements MethodBased
         // new optional parameters of constructors are not BC breaks,
         // as the method signature of child classes does not need to match the parent
         if ($fromMethod->isConstructor()) {
-            foreach ($added as $key => $paramName) {
-                $parameter = $toMethod->getParameter($paramName);
-                if ($parameter->isOptional()) {
-                    unset($added[$key]);
-                }
-            }
+            $added = array_filter($added, static fn (string $paramName) => !($toMethod->getParameter($paramName)?->isOptional()));
         }
 
         return Changes::fromList(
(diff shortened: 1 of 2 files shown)


src/DetectChanges/BCBreak/MethodBased/MethodParameterAdded.php
```diff
@@ -30,6 +30,17 @@ public function __invoke(ReflectionMethod $fromMethod, ReflectionMethod $toMetho
...
+        // new optional parameters of constructors are not BC breaks,
+        // as the method signature of child classes does not need to match the parent
+        if ($fromMethod->isConstructor()) {
+            foreach ($added as $key => $paramName) {
+                $parameter = $toMethod->getParameter($paramName);
+                if ($parameter->isOptional()) {
+                    unset($added[$key]);
+                }
+            }
+        }
```
Ocramius: can this be rewritten to be an `array_filter()` on `$toMethod->getParameters()`, please? 

As it is, this is a lot of added AST, and it therefore heavily weights on mutation tests and code complexity

Ocramius (approved):

src/DetectChanges/BCBreak/MethodBased/MethodParameterAdded.php
mstegmeyer: Yes, moved the filtering up and refactored it to an `array_filter`.

But: I decided to keep the outer `if` condition, since for most methods, that obviously aren't constructors, that filtering does not need to run and hog resources.

Ocramius (approved):
Thanks @mstegmeyer!

PR #991, opened 2026-03-20
chore: migrate to php-standard-library version 6 and drop php 8.3

azjezz:


src/Command/AssertBackwardsCompatible.php
```diff
@@ -175,8 +181,10 @@ public function execute(InputInterface $input, OutputInterface $output): int
                 $formatters[$format]->write($changes);
             }
         } finally {
-            $this->git->remove($fromPath);
-            $this->git->remove($toPath);
+            Async\concurrently([
+                fn () => $this->git->remove($fromPath),
+                fn () => $this->git->remove($toPath),
+            ]);
         }
```
azjezz, line 188: both call `Shell\execute` which waits, so can be called concurrently.

src/Command/AssertBackwardsCompatible.php
```diff
@@ -128,16 +132,18 @@ public function execute(InputInterface $input, OutputInterface $output): int
...
             Type\string()->coerce($fromRevision),
             Type\string()->coerce($toRevision),
         ));
 
-        $fromPath = $this->git->checkout($sourceRepo, $fromRevision);
-        $toPath   = $this->git->checkout($sourceRepo, $toRevision);
+        [$fromPath, $toPath] = Async\concurrently([
+            fn () => $this->git->checkout($sourceRepo, $fromRevision),
+            fn () => $this->git->checkout($sourceRepo, $toRevision),
+        ]);
```
azjezz, line 146: both call `Shell\execute` which waits, and do not rely on each other, can be ran concurrently.

Ocramius (approved):
🚢 thanks @azjezz!

PR #1005, opened 2026-09-16
fix: final internal property modification

### LINKED ISSUE #1004: Final classes do not respect "@internal" property annotation
We have a `final` class which has a public property marked with `@internal`. I would expect the breaking change detector to ignore any "breaking" changes to this property, as it does in a non-final class, but it ignores the `@internal` annotation when the class is final.

Steps to reproduce:

1. No `final` in class passes the backwards compatiblity check if the public property is `@internal` :

```diff
class MyClass
{
    /** @internal */
    public array $foo [
+        'bar',
    ];
}
```

```bash
$ ~/.composer/vendor/bin/roave-backward-compatibility-check
No backwards-incompatible changes detected
```

2. Classes marked `final` fail the backwards compatibility check even when marked `@internal`
```diff
final class MyClass
{
    /** @internal */
    public array $foo [
+        'bar',
    ];
}
```

```bash
$ ~/.composer/vendor/bin/roave-backward-compatibility-check
[BC] CHANGED: Property MyClass::$foo changed default value from array (
) to array (
  0 => 'foo',
)
```

bshaffer:
fixes #1004


Ocramius (approved):
Thanks @bshaffer!

